PR #14, opened 2020-03-16
Add the capability to observe subscription internal state

ocadaruma:
- Add the capability to observe subscription state change, which is useful for integration testing

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/SubscriptionStateListener.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/SubscriptionStateListener.java
@@ -38,6 +38,7 @@ import com.linecorp.decaton.processor.runtime.ProcessorSubscription;
  * }
  * </pre>
  */
+@FunctionalInterface
 public interface SubscriptionStateListener {
     /**
      * Represents possible states that a {@link ProcessorSubscription} can be in.
@@ -100,7 +101,7 @@ public interface SubscriptionStateListener {
     }
 
     /**
-     * Called after state transitioned to new state.
+     * Called at state transitioned to new state.
      * This method is called only from subscription thread (i.e. {@link ProcessorSubscription} itself)
      * @param newState new state of the subscription
      */
--- a/processor/src/main/java/com/linecorp/decaton/processor/SubscriptionStateListener.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/SubscriptionStateListener.java
@@ -43,36 +43,14 @@ public interface SubscriptionStateListener {
     /**
      * Represents possible states that a {@link ProcessorSubscription} can be in.
      * The expected state transition is:
-     *
      * <pre>
-     *
-     *                 +-----+--------+
-     *                 | INITIALIZING |
-     *                 +-----+--------+
-     *                       |
-     *                       v
-     *                 +--------------+
-     *                 | REBALANCING  |
-     *                 +----+--+------+
-     *                      |  ^
-     *                      v  |
-     *                 +----+--+------+
-     *                 |   RUNNING    |
-     *                 +-----+--------+
-     *                       |
-     *                       v
-     *                 +-----+--------+
-     *                 |SHUTTING_DOWN |
-     *                 +-----+--------+
-     *                       |
-     *                       v
-     *                 +-----+--------+
-     *                 | TERMINATED   |
-     *                 +--------------+
-     *
+     * {@code
+     * INITIALIZING -> REBALANCING <-> RUNNING -> SHUTTING_DOWN -> TERMINATED
+     * }
      * </pre>
+     *
      * Listener will be called at each transition in the flow.
-     * In addition, listener will be also called with INITIALIZING state when Decaton starts startup sequence
+     * In addition, listener will be also called with INITIALIZING state when Decaton starts initialization sequence.
      */
     enum State {
         /**
--- a/processor/src/main/java/com/linecorp/decaton/processor/SubscriptionStateListener.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/SubscriptionStateListener.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2020 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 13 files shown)


kawamuray:
Background;

* There was a demand for this feature to make it easy to synchronize state with decaton internal when writing integration tests (start producing after the subscription becomes ready)
* Another demand is to use this for healthcheking of subscriptions

processor/src/main/java/com/linecorp/decaton/processor/SubscriptionStateListener.java
```diff
@@ -0,0 +1,108 @@
...
+ * </pre>
+ */
+public interface SubscriptionStateListener {
+    /**
+     * Represents possible states that a {@link ProcessorSubscription} can be in.
+     * The expected state transition is:
+     *
+     * <pre>
+     *
+     *                 +-----+--------+
```
kawamuray: I think this diagram is not exactly correct since the transition from `RUNNING` to `SHUTTING_DOWN` is possibly to happen?

processor/src/main/java/com/linecorp/decaton/processor/SubscriptionStateListener.java
```diff
@@ -0,0 +1,108 @@
...
+         * Meanwhile no extra records are fetched until it completes.
+         */
+        REBALANCING,
+        /**
+         * Started to process tasks.
+         * Fetching records from brokers and feeding them into partition processors.
+         */
+        RUNNING,
+        /**
+         * Entered shutdown sequence.
```
kawamuray: `Entered` ?

processor/src/main/java/com/linecorp/decaton/processor/SubscriptionStateListener.java
```diff
@@ -0,0 +1,108 @@
...
+         */
+        SHUTTING_DOWN,
+        /**
+         * All shutdown sequence has done. The subscription is terminated completely.
+         */
+        TERMINATED,
+    }
+
+    /**
+     * Called after state transitioned to new state.
```
kawamuray: `after` is kind of ambiguous as in there's no associated "work" for each states? rather they're triggered at between two works. so "at"?

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -76,6 +80,17 @@ public ProcessorSubscription(SubscriptionScope scope,
         setName(String.format("DecatonSubscriptionThread-%s", scope));
     }
 
+    private void onStateChange(SubscriptionStateListener.State newState) {
+        logger.info("ProcessorSubscription transitioned to state: {}", newState);
+        if (stateListener != null) {
+            try {
+                stateListener.onChange(newState);
+            } catch (Exception e) {
+                logger.warn("State listener throws an exception", e);
```
kawamuray: "threw"

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -76,6 +80,17 @@ public ProcessorSubscription(SubscriptionScope scope,
         setName(String.format("DecatonSubscriptionThread-%s", scope));
     }
 
+    private void onStateChange(SubscriptionStateListener.State newState) {
```
kawamuray: since this is not a listener itself, `updateState` would be more appropriate naming?

processor/src/main/java/com/linecorp/decaton/processor/SubscriptionStateListener.java
```diff
@@ -0,0 +1,108 @@
...
+ *                        if (state == State.RUNNING) {
+ *                            latch.countDown();
+ *                        }
+ *                    }).buildAndStart();
+ * latch.await();
+ * // actual integration testing continues...
+ * }
+ * </pre>
+ */
+public interface SubscriptionStateListener {
```
kawamuray: `@FunctionalInterface`?

processor/src/test/java/com/linecorp/decaton/processor/runtime/ProcessorSubscriptionTest.java
```diff
@@ -0,0 +1,130 @@
...
+            pollLatch.countDown();
+            return records;
+        }).when(consumer).poll(anyLong());
+
+        ProcessorSubscription subscription = subscription(consumer, states::add, tp);
+
+        subscription.start();
+        pollLatch.await();
+
+        assertEquals(Arrays.asList(State.INITIALIZING,
```
kawamuray: image

same for below

processor/src/main/java/com/linecorp/decaton/processor/SubscriptionStateListener.java
ocadaruma: I omitted RUNNING to SHUTTING_DOWN because it only happens when something "unexpected" occurs e.g. OOOCC overflow.

....But, we know a case that ProcessorSubscription possibly die ungracefully when KafkaConsumer#committed throws an exception (which should be fixed later).
I'll add the transition.

processor/src/main/java/com/linecorp/decaton/processor/SubscriptionStateListener.java
ocadaruma: Oh? seems no difference?

processor/src/test/java/com/linecorp/decaton/processor/runtime/ProcessorSubscriptionTest.java
ocadaruma: Hmm, seems `expected, actual` is the correct order... (My IDE complains IF I flipped current order)

image

processor/src/main/java/com/linecorp/decaton/processor/SubscriptionStateListener.java
ocadaruma: RUNNING to SHUTTING_DOWN exists in the diagram.

REBALANCING to SHUTTING_DOWN right?

kawamuray: oops, no INITIALIZING to SHUTTING_DOWN.

processor/src/test/java/com/linecorp/decaton/processor/runtime/ProcessorSubscriptionTest.java
kawamuray: hmm, that's right, nevermind then.

processor/src/main/java/com/linecorp/decaton/processor/SubscriptionStateListener.java
ocadaruma: I imagined this diagram as usual expected transitions rather than all possible transitions, so I simplified the diagram.

kawamuray:
LGTM

PR #18, opened 2020-03-17
Add integration test fixes #17

### LINKED ISSUE #17: Add integration tests for decaton-processor
- Currently decaton-processor has no integration tests which interacts with real Kafka cluster
- The luck of integration tests makes refactoring or bug fix bit harder

ocadaruma:
## Summary
- Added `testing` project which contains embedded zk and kafka cluster
- integration test codes are located in `processor/src/it`
- Added a integration about retry-queuing feature


processor/src/it/java/com/linecorp/decaton/processor/RetryQueueingIntegration.java
```diff
@@ -0,0 +1,96 @@
...
+import com.linecorp.decaton.testing.KafkaClusterRule;
+
+public class RetryQueueingIntegration {
+    @ClassRule
+    public static KafkaClusterRule rule = new KafkaClusterRule();
+
+    @Test(timeout = 30000)
+    public void testRetryQueuing() throws Exception {
+        String topicName = "test-" + UUID.randomUUID();
+        rule.admin().createTopic(topicName, 3, 3);
```
kawamuray: Since those setup affects entire test cases that will be added to this class, it should be done in `@BeforeClass`.

processor/src/it/java/com/linecorp/decaton/processor/RetryQueueingIntegration.java
```diff
@@ -0,0 +1,96 @@
...
+                                 .thenProcess(processor),
+                RetryConfig.withBackoff(Duration.ofMillis(10)));
+             DecatonClient<HelloTask> client = TestUtils.client(topicName, rule.bootstrapServers())) {
+
+            keys.forEach(key -> client.put(key, HelloTask.getDefaultInstance()));
+            keysNeedRetry.forEach(key -> client.put(key, HelloTask.getDefaultInstance()));
+
+            processLatch.await();
+
+            assertEquals(keys, processedKeys);
```
kawamuray: no point to put those check in this try block?

processor/src/it/java/com/linecorp/decaton/processor/RetryQueueingIntegration.java
```diff
@@ -0,0 +1,96 @@
...
+        for (int i = 50; i < 100; i++) {
+            keys.add("key" + i);
+        }
+        Set<String> processedKeys = Collections.synchronizedSet(new HashSet<>());
+        Set<String> retriedKeys = Collections.synchronizedSet(new HashSet<>());
+        CountDownLatch processLatch = new CountDownLatch(keys.size() + keysNeedRetry.size());
+
+        DecatonProcessor<HelloTask> processor = (context, task) -> {
+            String key = context.key();
+            if (!key.endsWith("needRetry")) {
```
kawamuray: I think we'll add the another, most generic integration test that tests the core functionality of decaton (producer, consumer and process) which supposed to include those "normal process" cases right?
Would it be meaningful to test "non-retry" cases as well in this test case then?

processor/src/it/java/com/linecorp/decaton/processor/RetryQueueingIntegration.java
kawamuray: I'm guessing if it's good idea to test equity of two sets which contains 50 elements.
Let's say that this test fails, then it outputs the result of comparison for 50 elements, which is already difficult to glance by eyes but even if we could noticed that the particular key is missing here, would it be helps us to figure out the cause?

Having this test to fail, I imagine we'll anyway have to take some effort to enhance debugging information to and reproduce the issue several times so here it would be enough to compare just the size of two sets? 

Besides, I think we want to increase test cases to have much more than 50, as in one possible situation that we want to catch in integration test is race conditions, that makes more sense to have many items to test.

processor/src/it/java/com/linecorp/decaton/processor/TestUtils.java
```diff
@@ -0,0 +1,73 @@
...
+import com.linecorp.decaton.processor.runtime.SubscriptionBuilder;
+import com.linecorp.decaton.protobuf.ProtocolBuffersSerializer;
+
+public class TestUtils {
+
+    public static <T extends MessageLite> DecatonClient<T> client(String topic,
+                                                                  String bootstrapServers) {
+        Properties props = new Properties();
+        props.setProperty(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServers);
+        props.setProperty(ProducerConfig.CLIENT_ID_CONFIG, "test-client");
```
kawamuray: it's allowed to create multiple client instances in one test right? Then this should be differentiated at each call to prevent kafka library to complain client.id duplicates.

processor/src/it/java/com/linecorp/decaton/processor/TestUtils.java
```diff
@@ -0,0 +1,73 @@
...
+                            .producerConfig(props)
+                            .build();
+    }
+
+    public static <T> ProcessorSubscription subscription(String bootstrapServers,
+                                                         ProcessorsBuilder<T> processorsBuilder,
+                                                         RetryConfig retryConfig) {
+        Properties props = new Properties();
+        props.setProperty("bootstrap.servers", bootstrapServers);
+        props.setProperty("client.id", "test-processor");
```
kawamuray: likewise

processor/src/it/java/com/linecorp/decaton/processor/TestUtils.java
```diff
@@ -0,0 +1,73 @@
...
+import com.google.protobuf.MessageLite;
+
+import com.linecorp.decaton.client.DecatonClient;
+import com.linecorp.decaton.processor.SubscriptionStateListener.State;
+import com.linecorp.decaton.processor.runtime.ProcessorSubscription;
+import com.linecorp.decaton.processor.runtime.RetryConfig;
+import com.linecorp.decaton.processor.runtime.SubscriptionBuilder;
+import com.linecorp.decaton.protobuf.ProtocolBuffersSerializer;
+
+public class TestUtils {
```
kawamuray: I think this will be useful for any integration tests that might be implemented for different module in the future. How about moving this to testing module?

testing/src/main/java/com/linecorp/decaton/testing/EmbeddedKafkaCluster.java
```diff
@@ -0,0 +1,100 @@
...
+            String listener = "127.0.0.1:" + port;
+            listeners.add(listener);
+
+            log.info("Broker {} started at {}", i, listener);
+        }
+
+        bootstrapServers = String.join(",", listeners);
+    }
+
+    public String bootstrapServers() {
```
kawamuray: `@Getter`?

testing/src/main/java/com/linecorp/decaton/testing/EmbeddedZooKeeper.java
```diff
@@ -0,0 +1,78 @@
...
+
+        safeDelete(snapshotDir);
+        safeDelete(logDir);
+    }
+
+    private static void safeDelete(File file) {
+        try {
+            Utils.delete(file);
+        } catch (IOException e) {
+            log.info("Failed to delete {}", file, e);
```
kawamuray: likewise

testing/src/main/java/com/linecorp/decaton/testing/EmbeddedKafkaCluster.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @Override
+    public void close() {
+        for (KafkaServer server : servers) {
+            try {
+                server.shutdown();
+                server.awaitShutdown();
+            } catch (Exception e) {
+                log.info("Kafka broker {} threw an exception during shutting down", server.config().brokerId(), e);
```
kawamuray: not very important but `warn`? it's at least unexpected right?

processor/src/it/java/com/linecorp/decaton/processor/TestUtils.java
ocadaruma: Right, in the future, we possibly write a test which instantiates multiple producers in one test.
I'll fix.

processor/src/it/java/com/linecorp/decaton/processor/TestUtils.java
ocadaruma: Sounds make sense.

processor/src/it/java/com/linecorp/decaton/processor/TestUtils.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public static <T extends MessageLite> DecatonClient<T> client(String topic,
```
mauhiz: There should be a mechanism to ensure you don't reuse the same topic in different tests, since the `KafkaCluster` is not reset between tests

processor/src/it/java/com/linecorp/decaton/processor/RetryQueueingIntegration.java
ocadaruma: that's right. nice catch
> no point to put those check in this try block?

hm, that's completely enough
> compare just the size of two sets?

processor/src/it/java/com/linecorp/decaton/processor/RetryQueueingIntegration.java
ocadaruma: Okay, I'll add another test case which cover basic scenario.

kawamuray: In a separate PR?

We obviously want to test cases like start decaton apps -> rolling restart them -> make sure committed offsets are managed in expected way as a basic case, expecting it will involves certain volume of discussion.

ocadaruma: Hm, that's right.
So,  better to exclude RetryQueueing test from this PR as well ?

build.gradle
```diff
@@ -197,6 +210,6 @@ subprojects {
             archiveClassifier = 'javadoc'
             from javadoc
         }
-        build.dependsOn(javadoc)
+        check.dependsOn(javadoc, integrationTest)
```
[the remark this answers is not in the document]
ocadaruma: Hm right, integration tests will take time.
Okay, I'll run `integrationTest` task on CI explicitly.

processor/src/it/java/com/linecorp/decaton/processor/RetryQueueingIntegration.java
```diff
@@ -0,0 +1,96 @@
...
+import org.junit.Test;
+
+import com.linecorp.decaton.client.DecatonClient;
+import com.linecorp.decaton.processor.runtime.ProcessorSubscription;
+import com.linecorp.decaton.processor.runtime.RetryConfig;
+import com.linecorp.decaton.protobuf.ProtocolBuffersDeserializer;
+import com.linecorp.decaton.protocol.Sample.HelloTask;
+import com.linecorp.decaton.testing.KafkaClusterRule;
+
+public class RetryQueueingIntegration {
```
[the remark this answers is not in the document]
ocadaruma: IIRC there are some OSSs which suffixed with `Integration` in the wild.
But right, since we have separated task for integration tests we don't need to change suffix.
Let's go with `-Test`

processor/src/it/java/com/linecorp/decaton/processor/RetryQueueingIntegration.java
kawamuray: let's keep it. we need a concrete example to review helper stuff and it's already reviewed.

processor/src/it/java/com/linecorp/decaton/processor/RetryQueueingIntegration.java
ocadaruma: Means move to `@BeforeClass` and reuse same topic ?

processor/src/it/java/com/linecorp/decaton/processor/TestUtils.java
ocadaruma: Is random topic name (https://github.com/line/decaton/pull/18/files#diff-2f99db7e997743429863e8a0538f07c2R44) not enough?

processor/src/it/java/com/linecorp/decaton/processor/RetryQueueingTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+public class RetryQueueingTest {
+    @ClassRule
+    public static KafkaClusterRule rule = new KafkaClusterRule();
+
+    private String topicName;
+    private String retryTopicName;
+
+    @Before
+    public void setup() {
```
kawamuray: I think `setUp` and `tearDown` conventionally?

testing/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+dependencies {
+    implementation project(':processor')
+    implementation project(':protobuf')
+
+    implementation "org.apache.kafka:kafka-clients:$kafkaVersion"
+    implementation "org.apache.kafka:kafka_2.12:$kafkaVersion"
+    implementation "org.apache.kafka:kafka-clients:$kafkaVersion:test"
+    implementation "org.apache.kafka:kafka_2.12:$kafkaVersion:test"
+
+    // zookeeper also brings slf4j-log4j12 and log4j dependencies
```
kawamuray: if we expect it to be there (I think so as I see log4j.properties) at running test, let's not rely on transitive dependencies.

exclude them from this dep and add direct dep with controlled version is preferred.

build.gradle
```diff
@@ -57,14 +58,26 @@ subprojects {
...
+
+    sourceSets.create('it') {
+        compileClasspath += sourceSets.main.output
+        runtimeClasspath += sourceSets.main.output
+    }
+
+    task integrationTest(type: Test) {
+        testClassesDirs = sourceSets.it.output.classesDirs
+        classpath = sourceSets.it.runtimeClasspath
+        shouldRunAfter test
```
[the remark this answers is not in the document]
ocadaruma: It comes from https://docs.gradle.org/current/userguide/java_testing.html#sec:configuring_java_integration_tests, but currently we are executing ut and it independently so no point I guess.
I'll remove

testing/build.gradle
ocadaruma: I included log4j to make debug easier when integration test fails.

Agree, I'll add dependency explicitly.

PR #19, opened 2020-03-17
 Deprecate use of KafkaConsumer#committed by locally cached committed offset

### LINKED ISSUE #16: Exception raised from `KafkaConsumer#committed` could terminate subscription thread
There was a report from a user that tells their decaton instance's subscription thread got terminated with logs: `ProcessorSubscription .. got exception while consuming, currently assigned: [(topics)...]`.

After some investigation I found that `KafkaConsumer#committed` seems to have changed its behavior from "never to throw exception with indefinite timeout" to "throw an exception after request.timeout.ms" in between 1.1.x and the latest.

Since the `#committed` call is not guarded with exception handling we have to fix it.

kawamuray:
fixed #16 

By deprecating use of `KafkaConsumer#committed`, the issue is likely fixed but I guarded `offsetCommit` aginst `TimeoutException` as well to avoid future repeat of the same situation.

3rd commit contains completely unrelated change that I found when looking at some code.

### DIFF
--- a/docs/src/main/java/com/linecorp/decaton/DocumentChecker.java
+++ b/docs/src/main/java/com/linecorp/decaton/DocumentChecker.java
@@ -24,6 +24,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -33,6 +34,7 @@ import java.util.regex.Pattern;
 
 import org.asciidoctor.Asciidoctor;
 import org.asciidoctor.ast.DocumentHeader;
+import org.asciidoctor.jruby.internal.IOUtils;
 
 import lombok.Value;
 import lombok.experimental.Accessors;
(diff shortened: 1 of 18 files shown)


processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -167,7 +159,11 @@ public void run() {
                 @Override
                 public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
                     waitForRemainingTasksCompletion(rebalanceTimeoutMillis.value());
-                    commitCompletedOffsets(consumer);
+                    try {
+                        commitCompletedOffsets(consumer);
+                    } catch (CommitFailedException | TimeoutException e) {
+                        logger.warn("Offset commit failed at group rebalance", e);
```
ocadaruma: [nits] better to mention that process will be duplicated I think

docs/src/main/java/com/linecorp/decaton/DocumentChecker.java
```diff
@@ -176,6 +178,8 @@ public DocumentChecker() {
             }
 
             if (proc.exitValue() != 0) {
+                System.err.printf("git command %s failed with output: %s\n",
+                                  Arrays.toString(cmd), IOUtils.readFull(proc.getErrorStream()));
```
ocadaruma: nice

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
kawamuray: I think no, because of following reasons:

* Occurrence of this log doesn't mean there will be duplicate exactly, because everything's fine if it succeeds to commit the offset equals to or larger than the failed offset
* Explaining such complex situation like when will it be duplicate and when will not precisely in log isn't easy, as in I think too long log line is also not good.
* duplicate processing is widely recognized property of Kafka consumers, so I think users should be aware of such possibility at any cases, making it pointless to mention about it particularly here

kawamuray: > Occurrence of this log doesn't mean there will be duplicate exactly, because everything's fine if it succeeds to commit the offset equals to or larger than the failed offset on next round

nevermind about this. This is at rebalance listener so the duplicate will happen absolutely.

ocadaruma: agree with point 3 "duplicate processing is widely recognized property of Kafka consumers"

PR #30, opened 2020-03-30
Filter consumerConfig to use it as a producerConfig 

### LINKED ISSUE #29: Consumer's properties are also used for a producer to retry
The following type of warning occurs many times when using the retry mechanism of decaton.

```
The configuration 'group.id' was supplied but isn't a known config.
```

This warning occurs when a producer is given unused properties. 

https://github.com/apache/kafka/blob/779f1444f128e9d77486243cb94d7efcdea26f52/clients/src/main/java/org/apache/kafka/common/config/AbstractConfig.java#L353-L355

I found that the following code causes this warning.
https://github.com/line/decaton/blob/cd7cbbb54117f3df6b8ad52731b361372a0da552/processor/src/main/java/com/linecorp/decaton/processor/runtime/SubscriptionBuilder.java#L139-L140

Properties for a consumer are also used for properties for a retry producer.
I think this behavior is designed to set `bootstrap.servers` and so on without settings for retry.
But, more consumer's properties there are, the more warnings may occur.
It might be a good idea to filter consumer's properties in order to set the properties to a retry producer without warnings.

ta7uw:
Resolves #29 

`org.apache.kafka.clients.producer.ProducerConfig.configNames()` returns the set of producer properties names.  
Suppress warnings because of unused properties by taking an intersection of `consumerConfig` and it.

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/SubscriptionBuilder.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/SubscriptionBuilder.java
@@ -20,6 +20,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
+import java.util.function.Supplier;
 
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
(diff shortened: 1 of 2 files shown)


processor/src/main/java/com/linecorp/decaton/processor/runtime/SubscriptionBuilder.java
```diff
@@ -137,7 +139,18 @@ public ProcessorSubscription build() {
         DecatonProcessorSupplier<byte[]> retryProcessorSupplier = null;
         if (retryConfig != null) {
             Properties producerConfig = Optional.ofNullable(retryConfig.producerConfig())
-                                                .orElse(consumerConfig);
+                                                .orElseGet(() -> {
+                                                    Properties producerProps = new Properties();
+                                                    Set<String> definedProps = ProducerConfig.configNames();
+                                                    consumerConfig.entrySet()
```
kawamuray: I know it's weird but I think it is encouraged to not to rely on `Hashtable`'s methods when using `Properties` because most of them are not type-safe (`entrySet()` returns `Map<Object, Object>`). 
I think it's better to iterate over `stringPropertyNames()` with `get/setProperty` even for this trivial case.

ta7uw:
> Besides the inline comment, would you mind extracting your code into a package-private function and add a tiny unit test just in case?

I understand it.

kawamuray:
Apparently the pending build status is caused by an issue built the actual build has finished successfully.

PR #31, opened 2020-03-31
Fix use-after-close bug in DynamicRateLimiter

kawamuray:
## Explanation of bug

`DynamicRateLimiter` was using volatile variable to hold the currently active `RateLimiter` and replacing it with new one with new limit when it gets notified for the property update.
This was supposed to work well because users (calls `#acquire()`) might keep using old value read from `volatile current` which might have already rewritten by property listener.
However since property listener not just replaces reference but closes the limiter, this can cause race between property listener and hence user access already closed limiter consequently.

This could be reproduced by the unit test I added at c712cd6.

## Solution

The resulting solution became bit complex than I initially thought.
The patched version of `DynamicRateLimiter` makes use of both `volatile` var and RW lock to protect limiters against close while its being used.
In this solution, 
* readers throughput won't be affected because all they require is read lock which isn't exclusive with others. even when prop listener is updating the `current` limiter, it won't be block and can grab an active limiter just after 1 retry.
* writer can replace active limiter w/o lock. It has to wait on old limiter until all its users disappears to close the old limiter, but the time should be less and definite because all new entry of acquire() are routed to the new limiter.
See source comment for the detail.

Besides that, I have considered below (probably simpler) solutions but rejected them by individual reasons.

###  Depecate `RateLimiter#close`  with `RateLimiter#shutdown` and make it call optional
The idea of this solution is, by making RateLimiter's shutdown function call to be optional, when it got switched by property listener the user threads which grabbed the old value from `current` can just keep using the old one. This should be fine because currently all `RateLimiter` implementations implements `#close` just for immediately terminating all waiters on it.
However I turned out that by leaving (forgetting) the old `RateLimiter` in prop listener, it would causes extra long shutdown time of subscription by following scenario:
1.  thread 1..N grabs old limiter
2. prop listener replaces `current` with new one
3. subscription.close()
4. only the current limiter closed (so its waiters abort immediately)

### RW lock at DynamicRateLimiter
We might introduce RW lock at DynamicRateLimiter and acquire read for `#acquire()` calls while holding write for replacing `current` and close it.
This guarantees all users to grab a RateLimiter that is not already/going to be closed by prop listener.
However, the biggest concern is unexpected long lock either at users or at prop listener.

* If we make the lock's fair-policy "unfair", then while prop listener is awaiting on write lock, continuous calls of `acquire()` from users threads keeps interrupting and prevents prop listener to hold write lock indefinitely.
* If we make the lock's fair-policy "fair", then the above problem disappears, but instead some user threads will wait certainly and unnecessarily long while the write lock is waiting for other readers to release. In this case throughput might drops temporarily.

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/DynamicProperty.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/DynamicProperty.java
@@ -74,7 +74,7 @@ public class DynamicProperty<T> extends AbstractProperty<T> {
         validate(value);
 
         this.value = value;
-        logger.debug("Property {} has been updated ({} => {})", name(), currentValue, value);
+        logger.info("Property {} has been updated ({} => {})", name(), currentValue, value);
 
         notifyListeners(currentValue, value);
         return currentValue;
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/DynamicRateLimiter.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/DynamicRateLimiter.java
@@ -16,6 +16,8 @@
 
 package com.linecorp.decaton.processor.runtime;
 
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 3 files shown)


kawamuray:
The main concern of this patch has already been fixed by #32 as in RateLimiter no longer throws even for access after it closed.
I'll consider about this race once again and may submit a different PR to approach the remaining problem - `acquire()` returns too early and all at once at switch which is less critical than the original issue.

PR #32, opened 2020-04-02
Isolate exceptions threw from decaton's internal with those from user-supplied code

kawamuray:
This patch changes responsibility of `ProcessorUnit` and `ProcessPipeline`.
As-is:
* `ProcessorUnit` => manage per-thread queue, manage offset completes(especially, when processor returns w/o deferring completion) and some duration/error metrics 
* `ProcessPipeline` => task extraction, execution scheduling and some duration/error measurements

To-be:
* `ProcessorUnit` will focus on manage per-thread queue
* `ProcessPipeline` takes all the others and property distinguish failures at task extraction, scheduling and processing

All locations that pauses/sleeps upon task to become ready for processing are now checking termination flag so that once the termination starts they stops and returns immediately.

Now exceptions from user-supplied `DecatonProcessor` and unexpected exceptions from decaton's code are strictly distinguished and the latter causes consumption to stop by not comiting offset (should we terminate subscription in such case?)

All RateLimiters are made to not throw exception for access after close but returns immediately instead (and decaton skips the task accordingly to the termination flag).

PartitionProcessor's dependency termination order has fixed to regard child components inter-dependencies.

After this patch the existing special treatment against `InterruptedException` during shutdown is dropped because we found that it is pointless in typical implementation o processors which wraps entire `process()` logic with wide try-catch and handling in user's code. The reason it was only for `InterruptedException` is that it is one of the most typically thrown exception during shutdown but it lacks strong rationale since any other exceptions may be thrown instead of it (especially or libraries that prefer unchecked exceptions).

### DIFF
--- a/processor/src/it/java/com/linecorp/decaton/processor/RateLimiterTest.java
+++ b/processor/src/it/java/com/linecorp/decaton/processor/RateLimiterTest.java
@@ -18,7 +18,6 @@ package com.linecorp.decaton.processor;
 
 import static org.junit.Assert.assertEquals;
 
-import java.time.Duration;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
@@ -31,7 +30,6 @@ import org.junit.Test;
 
 import com.linecorp.decaton.client.DecatonClient;
 import com.linecorp.decaton.processor.runtime.ProcessorSubscription;
-import com.linecorp.decaton.processor.runtime.RetryConfig;
 import com.linecorp.decaton.protobuf.ProtocolBuffersDeserializer;
 import com.linecorp.decaton.protocol.Sample.HelloTask;
 import com.linecorp.decaton.testing.KafkaClusterRule;
@@ -42,18 +40,15 @@ public class RateLimiterTest {
     public static KafkaClusterRule rule = new KafkaClusterRule();
 
     private String topicName;
-    private String retryTopicName;
 
     @Before
     public void setUp() {
         topicName = rule.admin().createRandomTopic(3, 3);
-        retryTopicName = topicName + "-retry";
-        rule.admin().createTopic(retryTopicName, 3, 3);
     }
 
     @After
     public void tearDown() {
-        rule.admin().deleteTopics(topicName, retryTopicName);
+        rule.admin().deleteTopics(topicName);
     }
 
     @Test(timeout = 30000)
@@ -76,7 +71,7 @@ public class RateLimiterTest {
                 rule.bootstrapServers(),
                 ProcessorsBuilder.consuming(topicName, new ProtocolBuffersDeserializer<>(HelloTask.parser()))
                                  .thenProcess(processor),
-                RetryConfig.withBackoff(Duration.ofMillis(10)),
+                null,
                 StaticPropertySupplier.of(rateProp));
              DecatonClient<HelloTask> client = TestUtils.client(topicName, rule.bootstrapServers())) {
 
--- a/processor/src/it/java/com/linecorp/decaton/processor/RateLimiterTest.java
+++ b/processor/src/it/java/com/linecorp/decaton/processor/RateLimiterTest.java
@@ -37,7 +37,7 @@ import com.linecorp.decaton.protocol.Sample.HelloTask;
 import com.linecorp.decaton.testing.KafkaClusterRule;
 import com.linecorp.decaton.testing.TestUtils;
 
-public class RetryQueueingTest {
+public class RateLimiterTest {
     @ClassRule
     public static KafkaClusterRule rule = new KafkaClusterRule();
 
@@ -57,12 +57,9 @@ public class RetryQueueingTest {
     }
 
     @Test(timeout = 30000)
-    public void testRetryQueuing() throws Exception {
+    public void testPropertyDynamicSwitch() throws Exception {
         Set<String> keys = new HashSet<>();
 
-        // scenario:
-        //   * produce 10000 tasks.
-        //   * all tasks will be retried once, and processed after retried
         for (int i = 0; i < 10000; i++) {
             keys.add("key" + i);
         }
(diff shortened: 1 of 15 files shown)


ocadaruma:
build failing?

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessPipeline.java
```diff
@@ -60,28 +60,49 @@ public ProcessPipeline(ThreadScope scope,
...
+            log.debug("Returning after schedule due to termination");
+            return;
+        }
 
-        return process(request, extracted);
+        CompletableFuture<Void> result = process(request, extracted);
+        DeferredCompletion completion = request.completion();
+        result.whenComplete((r, e) -> {
+            if (e != null) {
+                if (e instanceof InterruptedException && true /* terminated */) {
```
ocadaruma, line 83: `&& true` ?

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessPipeline.java
```diff
@@ -60,28 +60,49 @@ public ProcessPipeline(ThreadScope scope,
...
-    public CompletableFuture<Void> scheduleThenProcess(TaskRequest request) throws InterruptedException {
-        DecatonTask<T> extracted = extract(request);
+    public void scheduleThenProcess(TaskRequest request) throws InterruptedException {
+        final DecatonTask<T> extracted;
+        try {
+            extracted = extract(request);
+        } catch (RuntimeException e) {
+            log.error("Dropping not-deserializable task [{}]", request.id(), e);
+            taskMetrics.tasksDiscarded.increment();
+            return;
```
ocadaruma, line 70: Isn't it necessary to call `completion.complete()` ?
IIUC Decaton shouldn't stop processing due to invalid tasks

https://github.com/line/decaton/blob/master/docs/consuming-any-data.adoc
image

processor/src/main/java/com/linecorp/decaton/processor/runtime/AveragingRateLimiter.java
```diff
@@ -55,14 +55,13 @@
...
     public long acquire(int permits) throws InterruptedException {
-        if (latch.getCount() == 0) {
-            throw new IllegalStateException("This limiter is already closed");
-        }
-
         long microsToWait = reserve(permits);
         if (microsToWait > 0L) {
             latch.await(microsToWait, MICROSECONDS);
         }
+        if (terminated()) {
```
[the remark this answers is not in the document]
kawamuray: It's a spec declared in interface 
image

The reason I thought that spec is better than returning partially awaited time is because those return values typically used to record metrics/logs of how long did it actually blocked, which is pointless or cases after `#close()` because callers are supposed to return immediately without proceeding to the next step after the limiter returns.

processor/src/main/java/com/linecorp/decaton/processor/runtime/ExecutionScheduler.java
```diff
@@ -60,27 +59,45 @@ void sleep(long millis) throws InterruptedException {
...
+    long waitOnScheduledTime(TaskMetadata metadata) throws InterruptedException {
+        long timeToWaitMs = metadata.scheduledTimeMillis() - currentTimeMillis.get();
+        if (timeToWaitMs > 0) {
+            log.debug("Sleeping {} ms awaiting scheduled time for task in {} - ({})",
+                      timeToWaitMs, scope, metadata);
+            sleep(timeToWaitMs);
         }
-
-        metrics.tasksSchedulingDelay.record(Math.max(timeUntilProcess, 0), TimeUnit.MILLISECONDS);
+        return timeToWaitMs;
```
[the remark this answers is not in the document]
kawamuray: right, let me fix with below unnecessary max for acquire as wel

processor/src/main/java/com/linecorp/decaton/processor/runtime/ExecutionScheduler.java
```diff
@@ -60,27 +59,45 @@ void sleep(long millis) throws InterruptedException {
...
+     * This method might returns earlier than the time ready to execute the task when {@link #close()} called.
+     *
+     * @param metadata the {@link TaskMetadata} associated with the task waiting to be processed.
+     * @throws InterruptedException when interrupted.
+     */
     public void schedule(TaskMetadata metadata) throws InterruptedException {
         // For tasks which are configured to delay its execution.
-        waitOnScheduledTime(metadata);
+        long timeWaitedMs = waitOnScheduledTime(metadata);
+        if (terminated()) {
```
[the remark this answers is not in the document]
kawamuray: Yes, because of:

1. It should not be recorded in below metric
2. It's pointless to enter rateLimiter which potentially blocks and is still unknown when exactly it will be closed despite we know the termination has began at this moment.

processor/src/main/java/com/linecorp/decaton/processor/runtime/RateLimiter.java
```diff
@@ -24,15 +24,39 @@
     long PAUSED = 0L;
     long MAX_RATE = SECONDS.toMicros(1L);
 
+    /**
+     * Acquire single execution from this limiter.
+     * Blocks until next execution is permitted and returns duration in microseconds that it has blocked.
```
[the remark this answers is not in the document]
kawamuray: oops, nice catch. forgot to fix the doc while I'm try-and-error around some different interfaces.

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessPipeline.java
kawamuray: as I mentioned briefly, I was still considering whether this special treatment worth it to keep or not. I'll make a decision and fix it soon.

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessPipeline.java
kawamuray: doh! super nice catch, thanks so much!
I'll add a test case to cover it and fix it to handle completion.

processor/src/test/java/com/linecorp/decaton/processor/runtime/ProcessPipelineTest.java
```diff
@@ -89,65 +98,134 @@ private static TaskRequest taskRequest() {
...
+            }
+        });
+        atSchedule.await();
+        pipeline.close();
+        closeLatch.countDown();
+
+        executor.shutdown();
+        // Checking it actually returns
+        executor.awaitTermination(Long.MAX_VALUE, TimeUnit.SECONDS);
+        verify(pipeline, never()).process(any(), any());
```
ocadaruma, line 228: ProcessPipeline is now responsible for managing offset completion, so better to check competion.compete() is not called ?

processor/src/main/java/com/linecorp/decaton/processor/runtime/AveragingRateLimiter.java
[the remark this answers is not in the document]
ocadaruma: understood.

processor/src/test/java/com/linecorp/decaton/processor/runtime/ProcessPipelineTest.java
kawamuray: good point, let me add that.

processor/src/it/java/com/linecorp/decaton/processor/RateLimiterTest.java
```diff
@@ -0,0 +1,95 @@
...
+            processedKeys.add(context.key());
+            processLatch.countDown();
+        };
+
+        DynamicProperty<Long> rateProp = new DynamicProperty<>(ProcessorProperties.CONFIG_PROCESSING_RATE);
+        try (ProcessorSubscription subscription = TestUtils.subscription(
+                rule.bootstrapServers(),
+                ProcessorsBuilder.consuming(topicName, new ProtocolBuffersDeserializer<>(HelloTask.parser()))
+                                 .thenProcess(processor),
+                RetryConfig.withBackoff(Duration.ofMillis(10)),
```
ocadaruma: [nits] Is that necessary to configure retry? As this test should focus on rate limit feature

PR #35, opened 2020-04-18
(in progress) Update HW before shutdown to reduce duplication fixes #23

### LINKED ISSUE #23: ProcessorSubscription should update HW in shutdown sequence

ocadaruma:
## Summary
- Update HW before shutdown (#23)

- Please start review after #47 got merged

### DIFF
--- a/testing/src/main/java/com/linecorp/decaton/testing/processor/ProcessorTestSuite.java
+++ b/testing/src/main/java/com/linecorp/decaton/testing/processor/ProcessorTestSuite.java
@@ -40,9 +40,9 @@ import lombok.experimental.Accessors;
 import lombok.extern.slf4j.Slf4j;
 
 /**
- * Test suite that checks Decaton's core functionality, such like processing semantics.
+ * Test suite that checks Decaton's core functionality, especially for processing semantics.
  *
- * A test will perform following procedure:
+ * Test suite will perform following procedure:
  *   1. Start multiple subscription instances
  *   2. Produce tasks
  *   3. When half of tasks are processed, restart subscriptions one by one
@@ -197,7 +197,7 @@ public class ProcessorTestSuite {
      * Generate and produce {@link #NUM_TASKS} tasks
      * @param producer Producer instance to be used
      * @param topic Topic to be sent tasks
-     * @param onProduce Callback which is called when a task is produced (!= completed to send)
+     * @param onProduce Callback which is called when a task is complete to be sent
      * @return A CompletableFuture of Map, which holds partition as the key and max offset as the value
      */
     private static CompletableFuture<Map<Integer, Long>> produceTasks(
@@ -230,11 +230,14 @@ public class ProcessorTestSuite {
             producer.send(record, (metadata, exception) -> {
                 if (exception == null) {
                     future.complete(metadata);
+                    onProduce.accept(new ProducedRecord(key,
+                                                        new TopicPartition(metadata.topic(), metadata.partition()),
+                                                        metadata.offset(),
+                                                        task));
                 } else {
                     future.completeExceptionally(exception);
                 }
             });
-            onProduce.accept(new ProducedRecord(key, task));
         }
 
         return CompletableFuture.allOf(produceFutures).thenApply(notUsed -> {
--- a/testing/src/main/java/com/linecorp/decaton/testing/processor/ProcessorTestSuite.java
+++ b/testing/src/main/java/com/linecorp/decaton/testing/processor/ProcessorTestSuite.java
@@ -1,251 +1,132 @@
 package com.linecorp.decaton.testing.processor;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
-
-import java.util.ArrayList;
-import java.util.Collections;
+import java.util.EnumSet;
 import java.util.HashMap;
-import java.util.HashSet;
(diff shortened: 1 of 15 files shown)


testing/src/main/java/com/linecorp/decaton/testing/processor/ProcessorTestSuite.java
```diff
@@ -0,0 +1,318 @@
...
+            while(tasks.hasNext()) {
+                DecatonProducerRecord<T> record = tasks.next();
+                client.put(record.key(), record.task());
+                List<T> tasks = producedTasks.computeIfAbsent(record.key(), key -> new ArrayList<>());
+                tasks.add(record.task());
+            }
+
+            rollingRestartLatch.await();
+            for (int i = 0; i < numSubscriptionInstance; i++) {
+                subscriptions[i].close();
```
kawamuray, line 274: For easier investigation when test fails, I think we wanna have more event loggingsaround here

testing/src/main/java/com/linecorp/decaton/testing/processor/ProcessorTestSuite.java
```diff
@@ -0,0 +1,318 @@
...
+                subscriptions[i] = subscriptionSupplier.get();
+            }
+
+            processLatch.await();
+            for (int i = 0; i < numSubscriptionInstance; i++) {
+                subscriptions[i].close();
```
kawamuray, line 280: here too

testing/src/main/java/com/linecorp/decaton/testing/processor/ProcessorTestSuite.java
```diff
@@ -0,0 +1,318 @@
...
+
+        /**
+         * Configure tasks which are produced to the topic.
+         * Note that all tasks should be differentiated from others (i.e. equals() never be true between tasks)
+         * since message ordering will be verified by checking produced task equals to processed task one by one.
+         *
+         * @param numTasks number of tasks
+         * @param taskIterator iterator which generates tasks
+         */
+        public Builder<T> produce(int numTasks,
```
kawamuray, line 189: I've also considered what's the most suitable interface for this "integration test framework" and summarized stuff like this:

Input(parameters) needs to be customizable:
* How `ProcessorSubscription` is constructed
* What `DecatonProcessor` does (e.g, retry(), throw exception etc)
* List of semantics expected to be achieved  with given setup (e.g, ordering guarantee is required by default but is no longer valid when retry is enabled)

Checks (those should be assumed by default, but should be possible to opt-out as the setup confirms to relax):
* Process ordering based on keys
* Duplicating tasks
* Missing tasks
* Serial processing (no two tasks sharing the same key gets processed at the same time)
* and maybe more to add

Other than those things, including the number of tasks to generate or what to include in each tasks are often not a matter. Given that, I think this framework can generate tasks with arbitrary default numbers by itself. (It would be possibly good to leave a room to add `opaque` in the task though. e.g, `.taskData(T opaque)`.

So first of all, I don't see necessity of letting users of this framework to fully custom generated tasks.

In addition, I think it's better that each checks are encupslated into a interface like `interface Check { ... } ` which holds methods to get checked-in on each event (e.g, receive task) and make them possible to opt-out with interface like `Suite.builder().optOutChecks(Checks.DELIVERY_ORDER)` or so.

testing/src/main/java/com/linecorp/decaton/testing/processor/ProcessorTestSuite.java
ocadaruma: ProcessorSubscription's logs not enough?

kawamuray: Perhaps, but I think it's much easier to see at which point the problematic thing had happen with associated information on control sequence (restart, stop) that is a contextual information available only at here.

testing/src/main/java/com/linecorp/decaton/testing/processor/ProcessorTestSuite.java
ocadaruma: Fair enough.
> I don't see necessity of letting users of this framework to fully custom generated tasks

Sounds nice, but how `interface Check` should be?
> I think it's better that each checks are encupslated into a interface like interface Check { ... }

For example, I think "Serial processing" can be checked by observing each received task (by having per-key mutex e.g.), but to check "Missing tasks", we have to pass "produced tasks" to detect which task is missing.

kawamuray: I think it can be something like below, taking a form of observer for each events.
Then it can accumulate contexts internally and use it to check on `assertSemantic`. 
There might be duplicate in information stored in each check but it would be reasonable since each check may requires different data structure to check against them eventually.

```java
    interface Check {
        void onProduce(String key, Task task);

        void onProcess(String key, Task task);

        void assertSemantic();
    }
```


Regarding "Serial processing" I think it's better to record process being/end time of each records and confirm there're no overlap between two records sharing the same key.

ocadaruma: Much simpler, looks good
> confirm there're no overlap between two records

ocadaruma:
I'll recreate another PR

PR #36, opened 2020-04-23
Add profiling capability to benchmark

kawamuray:
## Motivation

The benchmark module is useful not only for tracking performance regression between decaton versions but also for persuading even further performance optimizations and/or to figure out why the performance regression has happened.
To get more insights from benchmark we need more information from it most significantly profiling.
Of course we can run profilers like async-profiler externally with synchronizing timing with benchmark execution, but it's highly useful to integrate the profiling into benchmark itself so that:
* We can easily perform profiling just by single extra command line switch
* Benchmark code knows more about when it should start profiling. e.g, should not profile until warmup completes.

## Major Changes

* Made benchmark runner to fork a separate JVM to run consumer instance. This has several reasons like 1. to avoid GC/allocation workload from unrelated thing like embedded kafka cluster, producers, 2. to remove noises from profiling result: Related classes: `Execution`, `InProcessExecution`, `ForkingExecution`
* Add async-profiler support. Related classes: `Profiling`, `AsyncProfilerProfiling`
* Add JVM (GC) stats in output. Related classes: `JvmTracker`
* Add some metrics to observe partition processors' queue starving status. Related classes: `Metrics`
* Wait until JIT compilation to get settled after warmup before proceeding to real run. Related classes: `Benchmark`

Example output:
```
=== Decaton (100000 tasks) ===
# Runner: com.linecorp.decaton.benchmark.DecatonRunner
# Tasks: 100000 (warmup: 100000)
# Simulated Latency(ms): 0
# Param: decaton.partition.concurrency=20
--- Performance ---
Execution Time(ms): 4569.00
Throughput: 21886.63 tasks/sec
Delivery Latency(ms): mean=954 max=3049
--- Resource Usage (61 threads observed) ---
Cpu Time(ms): 9071.49
Allocated Heap(KiB): 831205.57
--- JVM ---
GC (G1 Old Generation) Count: 0
GC (G1 Old Generation) Time(ms): 0
GC (G1 Young Generation) Count: 5
GC (G1 Young Generation) Time(ms): 45
```

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -13,11 +13,11 @@ Build a shadowJar:
 
 Run with arbitrary parameters:
 ```sh
-java -cp build/libs/benchmark-VERSION-shadow.jar com.linecorp.decaton.benchmark.Main \
+./debm.sh
  --title "Decaton" \
  --runner com.linecorp.decaton.benchmark.DecatonRunner \
  --tasks 10000 \
  --warmup 100 \
- --max-latency=10 \
+ --simulate-latency=10 \
  --param=decaton.partition.concurrency=20 
 ```
--- a/benchmark/src/main/java/com/linecorp/decaton/benchmark/AsyncProfilerProfiling.java
+++ b/benchmark/src/main/java/com/linecorp/decaton/benchmark/AsyncProfilerProfiling.java
@@ -95,7 +95,11 @@ public class AsyncProfilerProfiling implements Profiling {
     private Optional<Path> findOutputPath() {
         for (int i = 0; i < asyncProfilerOpts.size() - 1; i++) {
             if ("-f".equals(asyncProfilerOpts.get(i))) {
-                return Optional.of(Paths.get(asyncProfilerOpts.get(i + 1)));
+                // Avoid printing absolute path expecting the result to be copy & pasted in public
+                // place by security concerns.
+                Path path = Paths.get("").toAbsolutePath().relativize(
+                        Paths.get("", asyncProfilerOpts.get(i + 1)));
+                return Optional.of(path);
             }
         }
         return Optional.empty();
--- a/benchmark/src/main/java/com/linecorp/decaton/benchmark/AsyncProfilerProfiling.java
+++ b/benchmark/src/main/java/com/linecorp/decaton/benchmark/AsyncProfilerProfiling.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright 2020 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.decaton.benchmark;
+
+import java.io.File;
+import java.lang.ProcessBuilder.Redirect;
+import java.lang.management.ManagementFactory;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Profiler for execution.
+ * Internally uses async-profiler https://github.com/jvm-profiling-tools/async-profiler to get
+ * stack samples.
+ */
+@Slf4j
+public class AsyncProfilerProfiling implements Profiling {
+    private static final int PROFILER_CMD_TIMEOUT_SECS = 30;
+
+    private final Path asyncProfilerBin;
+    private final List<String> asyncProfilerOpts;
+
+    public AsyncProfilerProfiling(Path asyncProfilerBin, Collection<String> asyncProfilerOpts) {
+        this.asyncProfilerBin = asyncProfilerBin;
+        if (asyncProfilerOpts == null) {
+            this.asyncProfilerOpts = new ArrayList<>();
+        } else {
+            this.asyncProfilerOpts = new ArrayList<>(asyncProfilerOpts);
+        }
+        if (!this.asyncProfilerOpts.contains("-f")) {
+            this.asyncProfilerOpts.add("-f");
+            this.asyncProfilerOpts.add(outputFileName());
+        }
+    }
+
+    private static long currentPid() {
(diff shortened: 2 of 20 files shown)


benchmark/src/main/java/com/linecorp/decaton/benchmark/InProcessExecution.java
```diff
@@ -0,0 +1,144 @@
...
+            }
+
+            profiling.start();
+            JvmTracker jvmTracker = JvmTracker.create();
+            stageCallback.accept(Stage.READY);
+            if (!recording.await(3, TimeUnit.MINUTES)) {
+                throw new RuntimeException("timeout on awaiting benchmark to complete");
+            }
+            profiling.stop().ifPresent(
+                    outputPath -> log.info("Profiling output is available at {}", outputPath));
```
ocadaruma, line 85: I think profiling output should be included in BenchmarkResult and be printed in [BenchmarkResult#print](https://github.com/line/decaton/pull/36/files#diff-a3e470608551e514a3285130e4b7b46dR135) because when we omit profiler output path, AsyncProfilerProfiling automatically determine output path and we have to find that from logs to see profile result.

benchmark/src/main/java/com/linecorp/decaton/benchmark/Main.java
```diff
@@ -30,28 +38,72 @@
...
 
+    @Option(names = "--profile", description = "Enable profiling of execution with async-profiler")
+    private boolean enableProfiling;
+
+    @Option(names = "--profiler-bin", description = "Path to async-profiler's profiler.sh",
+            defaultValue = "profiler.sh")
+    private Path profilerBin;
+
+    @Option(names = "--profiler-opts", description = "Options to pass for async-profiler's profiler.sh")
+    private String profilerOpts;
```
ocadaruma, line 74: I think awaiting JIT or not also should be cofigurable, because I found that awaiting JIT takes certainly long time when I run locally (2.5 min in my case).

When we encounter performance issue, we might change parameter or code and take profile several times, it can make the iteration slower.

benchmark/src/main/java/com/linecorp/decaton/benchmark/InProcessExecution.java
kawamuray: hm, that's a good point. 
The reason I didn't included it in the result is because I wanted to make `BenchmarkResult` portable, that is it might be serialized into JSON and brought around places like pasting the result making it not reasonable to contain contextual information like filepath.

but your point about the log is reasonble so, may I add a field to BenchmarkResult that doesn't gets serialized into JSON but yet printed in output?

benchmark/src/main/java/com/linecorp/decaton/benchmark/Main.java
kawamuray: hm, okay then let me add an option to disable it at least.

processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
```diff
@@ -44,6 +45,28 @@ private Metrics(AvailableTags availableTags) {
         this.availableTags = availableTags;
     }
 
+    public class SubscriptionMetrics {
+        private Timer processDuration(String scope) {
+            return Timer.builder("subscription.process.durations")
+                        .description(String.format(
+                                "Time spent for processing %s in consuming loop", scope))
+                        .tags(availableTags.subscriptionScope().and("scope", scope))
```
ocadaruma, line 53: - `scope` appears twice as different meaning here (subscription "scope" and consuming loop scope), so `part` sounds more appropriate

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -229,7 +233,9 @@ public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
                 pollOnce(consumer);
                 if (System.currentTimeMillis() - lastCommittedMillis >= commitIntervalMillis.value()) {
                     try {
+                        Timer timer = Utils.timer();
                         commitCompletedOffsets(consumer);
+                        metrics.commitOffsetTime.record(timer.duration());
```
ocadaruma, line 238: - This should be recorded outside of `if` block I think, because `SubscriptionMetrics` sounds like "the breakdown of each consuming loop" (as description of the metrics)
- And if we count this inside `if` block, I guess 99%ile and 99.9%ile value cannot be measured correctly since expiry is set to 60sec but commit interval is 1000ms by default

benchmark/README.md
```diff
@@ -13,11 +13,11 @@ Build a shadowJar:
 
 Run with arbitrary parameters:
 ```sh
-java -cp build/libs/benchmark-VERSION-shadow.jar com.linecorp.decaton.benchmark.Main \
+./debm.sh
```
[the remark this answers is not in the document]
kawamuray: oops, completely missed to stage the file;;

benchmark/src/main/java/com/linecorp/decaton/benchmark/AsyncProfilerProfiling.java
```diff
@@ -0,0 +1,114 @@
...
+        exec("start");
+    }
+
+    private Optional<Path> findOutputPath() {
+        for (int i = 0; i < asyncProfilerOpts.size() - 1; i++) {
+            if ("-f".equals(asyncProfilerOpts.get(i))) {
+                // Avoid printing absolute path expecting the result to be copy & pasted in public
+                // place by security concerns.
+                Path path = Paths.get("").toAbsolutePath().relativize(
+                        Paths.get("", asyncProfilerOpts.get(i + 1)));
```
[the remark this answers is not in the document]
kawamuray: That's true, shame I missed it in testing locally ;;
As the last resort I think I can make the field to have type `String` rather than `Path` so that it won't get converted into fullpath. I'll try that. 

> So instead of relativize here, necessary to construct relative path in BenchmarkResult#print?

I don't want to do this because I want to keep the same security property for (to be added) JSON output.

processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
```diff
@@ -103,6 +126,12 @@ private Metrics(AvailableTags availableTags) {
                           .tags(availableTags.partitionScope())
                           .register(registry);
 
+        public final Counter queueStarvedTime =
```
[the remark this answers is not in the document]
kawamuray: Oops, good catch, fixing.

processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
kawamuray: That's reasonable, but may I choose `section` instead? It feels more natural to me.

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
kawamuray: Ah that's very nice catch, yes indeed, fixing, thx.

PR #38, opened 2020-05-21
Cleanup all metrics at corresponding shutdown

kawamuray:
Currently once Decaton's internal metrics is created it never gets cleaned up. 
This is causing some metrics to leftover in scrape resulting making it very confusing to distinguish the value already became invalid by partition reassignment vs the currently active and value.
This PR ensures all metrics to cleanup at their corresponding scope's destruction time.

Also the micrometer dependency is upgraded to make the meter removal function work properly.

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
@@ -17,8 +17,11 @@
 package com.linecorp.decaton.processor.metrics;
 
 import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
 
 import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.Meter;
 import io.micrometer.core.instrument.Meter.Id;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Timer;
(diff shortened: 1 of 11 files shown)


processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
```diff
@@ -45,15 +52,55 @@ private Metrics(AvailableTags availableTags) {
...
+                              .incrementAndGet();
+            }
+            meters.add(meter);
+            return meter;
+        }
+
+        @Override
+        public void close() {
+            for (Meter meter : meters) {
+                meter.close();
```
ocadaruma, line 82: I'm not sure if it's safe to close the meter here, since the meter can be updated/queried after close until all references deleted.

As far as checking Meter impls in micrometer-core, it seems `close()` is not overridden in almost all cases so maybe it's safe to close here, but I think it's better to move `close()` to this line: https://github.com/line/decaton/pull/38/files#diff-78df474b860bfdbeae88ca6d9ebd1c17R88

kawamuray: That's a good point. Actually I've also considered which scope should we expect as a lifecycle for a meter instance.
My thought is, since `XXBuilder.build()` (which internally calls `registry.counter, gauge, timer...`) transparently returns an existing object if already presents, the returned objective's `close` should be safe to be called w/o extra assumption against possibly shared other instances.
On the other hand, `registry.remove` is difference since we're operating directly on the registry and explicitly ordering it to "remove" the given ID from it which it makes sense for me to work as a global destruction.
Currently the close method seems to be really pointless (not doing anything) except few cases like Timer as you said, so it really doesn't matter, but wouldn't the above interpretation makes much sense?

processor/src/main/java/com/linecorp/decaton/processor/metrics/ValueGauge.java
```diff
@@ -68,12 +87,12 @@ public static Builder builder(String name) {
...
         public ValueGauge register(MeterRegistry registry) {
             AtomicInteger underlying = new AtomicInteger(0);
-            Gauge.builder(name, underlying::get)
-                 .description(description)
-                 .baseUnit(baseUnit)
-                 .tags(tags)
-                 .register(registry);
-            return new ValueGauge(underlying);
+            Gauge gauge = Gauge.builder(name, underlying::get)
+                                  .description(description)
```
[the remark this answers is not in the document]
kawamuray: oops, thanks

kawamuray:
> Does it mean, with current version of micrometer, meter cannot be removed from registry ?

Yes, without upgrading micrometer version the newly added `MetricsCleanupTest` keeps failing with insisting the registry to still contain 42 meter instances even after subscription close.

processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
ocadaruma: Hm that makes sense. Thank you for your explanation

ocadaruma:
I see. thanks

processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
kawamuray: Hmm, after I looked into `AbstractTimer.close => PauseDetector.close` now I changed my mind that it would not be expected to call close multiple times from difference instances...
Let me fix in a way you suggested then.

PR #40, opened 2020-05-22
Replace decaton_partitions_paused metrics with per-partition boolean metric to make it work correctly

kawamuray:
* Currently `decaton_partitions_paused` metric is not working as expected because its actual counter held in `ValueGauge` is not shared among different `PartitionContext`s and only the first counter (first registered) value is shown as its value.
* Same thing happens to `decaton_tasks_pending` when rebalance occurs because the newly created ValueGauge doesn't actually gets registered.

To fix it, this PR:
* Deprecates ValueGauge and replace it with pure Gauge
* Adds integration test that checks metrics' stuck detection functionality


processor/src/main/java/com/linecorp/decaton/processor/runtime/PartitionContext.java
```diff
@@ -52,7 +52,7 @@ public PartitionContext(PartitionScope scope, Processors<?> processors, int maxP
         metrics = Metrics.withTags("subscription", scope.subscriptionId(),
                                    "topic", tp.topic(),
                                    "partition", String.valueOf(tp.partition()))
-                .new PartitionStateMetrics();
+                .new PartitionStateMetrics(commitControl::pendingOffsetsCount, () -> paused() ? 1 : 0);
```
kawamuray, line 55: I remember there was a discussion about letting metrics to call this `pendingOffsetsCount` directly with concern of race between consumer thread and metric scrape.
After I looked into its implementation, what it does is just calling .size() against ArrayDeque which should return ultimately fast and since other synchronized methods call by consumer thread are just tiny part of entire consuming-loop I thought it would be safe here. Different thoughts? @ocadaruma

ocadaruma: Sounds completely fine.

processor/src/it/java/com/linecorp/decaton/processor/MetricsTest.java
```diff
@@ -70,4 +75,39 @@ public void testMetricsCleanup() throws Exception {
         List<Meter> meters = Metrics.registry().getMeters();
         assertEquals(emptyList(), meters);
     }
+
+    @Test(timeout = 30000)
+    public void testDetectStuckPartitions() throws Exception {
+        // Micrometer doesn't track values without at least one register implementation added
+        Metrics.register(new PrometheusMeterRegistry(PrometheusConfig.DEFAULT));
+
+        CountDownLatch processLatch = new CountDownLatch(PARTITIONS * 2);
```
ocadaruma, line 84: just `new CountDownLatch(PARTITIONS)` not work ?

processor/src/it/java/com/linecorp/decaton/processor/MetricsTest.java
```diff
@@ -70,4 +75,39 @@ public void testMetricsCleanup() throws Exception {
...
+                        Property.ofStatic(ProcessorProperties.CONFIG_PARTITION_CONCURRENCY, 1),
+                        Property.ofStatic(ProcessorProperties.CONFIG_MAX_PENDING_RECORDS, 1)));
+             DecatonClient<HelloTask> client = TestUtils.client(topicName, rule.bootstrapServers())) {
+            for (int i = 0; i < PARTITIONS * 10; i++) {
+                // What we need here is just simple round-robin but it's not possible with null keys anymore
+                // since https://cwiki.apache.org/confluence/display/KAFKA/KIP-480%3A+Sticky+Partitioner
+                client.put(String.valueOf(i), HelloTask.getDefaultInstance());
+            }
+            processLatch.await();
+            Thread.sleep(1000);
```
ocadaruma, line 104: What's the purpose of this sleep ?

processor/src/it/java/com/linecorp/decaton/processor/MetricsTest.java
```diff
@@ -70,4 +75,39 @@ public void testMetricsCleanup() throws Exception {
...
+        try (ProcessorSubscription subscription = TestUtils.subscription(
+                rule.bootstrapServers(),
+                ProcessorsBuilder.consuming(topicName, new ProtocolBuffersDeserializer<>(HelloTask.parser()))
+                                 .thenProcess(processor),
+                null,
+                StaticPropertySupplier.of(
+                        Property.ofStatic(ProcessorProperties.CONFIG_PARTITION_CONCURRENCY, 1),
+                        Property.ofStatic(ProcessorProperties.CONFIG_MAX_PENDING_RECORDS, 1)));
+             DecatonClient<HelloTask> client = TestUtils.client(topicName, rule.bootstrapServers())) {
+            for (int i = 0; i < PARTITIONS * 10; i++) {
```
ocadaruma, line 98: What's the rationale of `* 10` ?

processor/src/it/java/com/linecorp/decaton/processor/MetricsTest.java
kawamuray: so after very tough try & fails finally I've managed to remove all those flacky-ish stuff from this test, along with fixing core to regard `decaton.max.pending.records` at exact boundary: cbc7d75 

it could still be failed on race scenario like this test confirms paused count before subscription thread to reach `pausePartitions` but its possibility is likely very low. we cannot await on next records arrival either because all partitions are already paused and processors won't get any of them.

kawamuray:
The above test failure is by performance bound.

PR #41, opened 2020-05-25
Improve comment for deferred completion when using retry

skkap:
According to this documentation: https://github.com/line/decaton/blob/master/docs/retry-queueing.adoc#retry-with-asynchronous-processing and implementation deferred completion will be complete when `retry()` is called. 
The current comment is confusing as it says 
```
caller *MUST* call {@link DeferredCompletion#complete()} method in any cases.
```
which is not true with the current implementation.


kawamuray:
LGTM, thanks!

kawamuray:
Build failure seems to be not related the change (performance bound)

PR #42, opened 2020-05-26
Fix flaky testDetectStuckPartitions to be stable

kawamuray:
Fixup for flaky integration test added by #40


ocadaruma:
Confirmed that failing test is performance bound test

PR #44, opened 2020-05-26
Revert PerformanceBoundTest to stop flacky test running temporarily

kawamuray:
Since the test is too flaky due to large variance of performance on travis CI, I'm gonna disable it once and consider much stabler way to continuously measure decaton's performance.


processor/src/it/java/com/linecorp/decaton/processor/PerformanceBoundTest.java
```diff
@@ -1,93 +0,0 @@
...
-import org.junit.Assume;
-import org.junit.Before;
-import org.junit.Test;
-
-import com.linecorp.decaton.benchmark.Benchmark;
-import com.linecorp.decaton.benchmark.BenchmarkConfig;
-import com.linecorp.decaton.benchmark.BenchmarkResult;
-import com.linecorp.decaton.benchmark.DecatonRunner;
-
-public class PerformanceBoundTest {
```
mauhiz, line 36: just `@Ignore` it?

kawamuray: That also works, but since I'm not sure if we'll keep this way of testing performance (run as part of integration test), I wanted to clean things up once and add it back later if necessary to avoid leaving dead code.

PR #48, opened 2020-06-11
Update HW in shutdown procedure to reduce duplication fixes #23

### LINKED ISSUE #23: ProcessorSubscription should update HW in shutdown sequence

ocadaruma:
## Summary
- Update H/W before shutting down to reduce process duplication
- By this fix, as long as we set partition.concurrency to 1 as like vanilla consumer, we can expect process duplication not to occur.
  * Added a test about that

### DIFF
--- a/processor/src/it/java/com/linecorp/decaton/processor/CoreFunctionalityTest.java
+++ b/processor/src/it/java/com/linecorp/decaton/processor/CoreFunctionalityTest.java
@@ -16,6 +16,12 @@
 
 package com.linecorp.decaton.processor;
 
+import static org.junit.Assert.assertTrue;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.Random;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -27,7 +33,12 @@ import org.junit.Test;
 import com.linecorp.decaton.processor.runtime.ProcessorScope;
 import com.linecorp.decaton.testing.KafkaClusterRule;
 import com.linecorp.decaton.testing.RandomRule;
+import com.linecorp.decaton.testing.processor.ProcessedRecord;
+import com.linecorp.decaton.testing.processor.ProcessingGuarantee;
+import com.linecorp.decaton.testing.processor.ProcessingGuarantee.GuaranteeType;
 import com.linecorp.decaton.testing.processor.ProcessorTestSuite;
+import com.linecorp.decaton.testing.processor.ProducedRecord;
+import com.linecorp.decaton.testing.processor.TestTask;
 
 public class CoreFunctionalityTest {
     @ClassRule
@@ -100,6 +111,51 @@ public class CoreFunctionalityTest {
                         }
                     });
                 }))
+                .build()
+                .run();
+    }
+
+    /**
+     * Decaton doesn't guarantee that produced messages are processed in
+     * "strictly" same order (refs {@link GuaranteeType#PROCESS_ORDERING}).
+     *
+     * But as long as we set partition.concurrency to 1 and all subscriptions are
+     * restarted gracefully, we can expect "strict" process ordering as like vanilla Kafka consumer
+     */
+    @Test(timeout = 60000)
+    public void testSingleThreadProcessing() {
+        ProcessingGuarantee strictOrdering = new ProcessingGuarantee() {
+            private final Map<String, List<TestTask>> produced = new HashMap<>();
+            private final Map<String, List<TestTask>> processed = new HashMap<>();
+
+            @Override
+            public synchronized void onProduce(ProducedRecord record) {
+                produced.computeIfAbsent(record.key(), key -> new ArrayList<>()).add(record.task());
+            }
+
+            @Override
+            public synchronized void onProcess(TaskMetadata metadata, ProcessedRecord record) {
+                processed.computeIfAbsent(record.key(), key -> new ArrayList<>()).add(record.task());
+            }
+
+            @Override
+            public void doAssert() {
+                // use assertTrue instead of assertEquals not to cause error message explosion
+                //noinspection SimplifiableJUnitAssertion
+                assertTrue(produced.equals(processed));
+            }
+        };
+
+        Random rand = randomRule.random();
+        ProcessorTestSuite
+                .builder(rule)
+                .configureProcessorsBuilder(builder -> builder.thenProcess((ctx, task) -> {
+                    Thread.sleep(rand.nextInt(10));
+                }))
+                .propertySupplier(StaticPropertySupplier.of(
+                        Property.ofStatic(ProcessorProperties.CONFIG_PARTITION_CONCURRENCY, 1)
+                ))
+                .customSemantics(strictOrdering)
                 .build()
                 .run();
     }
--- a/processor/src/it/java/com/linecorp/decaton/processor/CoreFunctionalityTest.java
+++ b/processor/src/it/java/com/linecorp/decaton/processor/CoreFunctionalityTest.java
@@ -35,7 +35,6 @@ import com.linecorp.decaton.testing.KafkaClusterRule;
 import com.linecorp.decaton.testing.RandomRule;
 import com.linecorp.decaton.testing.processor.ProcessedRecord;
 import com.linecorp.decaton.testing.processor.ProcessingGuarantee;
-import com.linecorp.decaton.testing.processor.ProcessingGuarantee.GuaranteeType;
 import com.linecorp.decaton.testing.processor.ProcessorTestSuite;
 import com.linecorp.decaton.testing.processor.ProducedRecord;
 import com.linecorp.decaton.testing.processor.TestTask;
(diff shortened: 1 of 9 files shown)


processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -244,6 +249,7 @@ public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
             Timer timer = Utils.timer();
             contexts.destroyAllProcessors();
             try {
+                contexts.updateHighWatermarks();
```
kawamuray, line 252: Any points to put this in inside try block while we don't expect it to throw?

kawamuray: Let's leave a comment describing this situation more like:
"Since kafka version xx the rebalance listener gets triggered at close so the same thing is supposed to happen, but its not true for older kafka-clients version so we're manually handling this here instead of relying on rebalance listener for better compatibility" something like this.

processor/src/it/java/com/linecorp/decaton/processor/CoreFunctionalityTest.java
```diff
@@ -103,4 +114,49 @@ public void testAsyncTaskCompletion() {
                 .build()
                 .run();
     }
+
+    /**
+     * Decaton doesn't guarantee that produced messages are processed in
```
kawamuray, line 119: To avoid misleading, it would better to clarify state that
- this is not a part of spec
- however, we as developer can expect this behavior when setting concurrency to 1 so just checking that expectation here

processor/src/it/java/com/linecorp/decaton/processor/CoreFunctionalityTest.java
```diff
@@ -103,4 +114,49 @@ public void testAsyncTaskCompletion() {
...
+     * "strictly" same order (refs {@link GuaranteeType#PROCESS_ORDERING}).
+     *
+     * But as long as we set partition.concurrency to 1 and all subscriptions are
+     * restarted gracefully, we can expect "strict" process ordering as like vanilla Kafka consumer
+     */
+    @Test(timeout = 60000)
+    public void testSingleThreadProcessing() {
+        ProcessingGuarantee strictOrdering = new ProcessingGuarantee() {
```
kawamuray, line 127: Is strictOrdering` what we expect for this? I thought it's something more a like "no duplicates"?

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -179,6 +179,11 @@ public void run() {
             consumer.subscribe(subscribeTopics(), new ConsumerRebalanceListener() {
                 @Override
                 public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
+                    // KafkaConsumer#close has been changed to invoke onPartitionRevoked since Kafka 2.4.0.
+                    // Since we're doing cleanup procedure on shutdown individually so just immediately return if terminated
+                    if (terminated.get()) {
```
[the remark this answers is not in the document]
ocadaruma: Hmm sounds true.... thanks for pointing out.

Instead of adding another flag like `isClosing`, how about storing subscription's state in `updateState` and checking it? (`if (state == State.SHUTTING_DOWN) {...`)

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
ocadaruma: Hm, no.
I'll move to outside of try block.
> Any points to put this in inside try block

processor/src/it/java/com/linecorp/decaton/processor/CoreFunctionalityTest.java
ocadaruma: thanks, `noDuplicates` sounds more appropriate.

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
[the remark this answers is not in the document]
kawamuray: Sounds workable.
We would have few concerns thought;
- As of current version kafka states that the rebalance listener is invoked only during active call of `Consumer#poll`(see javadoc of `KafkaConsumer#subscribe`, which is already a tiny lie since it also gets invoked during `Consumer#close`....) so it should be safe to use `SHUTTING_DOWN` state flag as-is. However if in future, either we add a new operation in between `updateState(SHUTTING_DOWN)` and `consumer.close()`, we would have to introduce another State variant which is pointless for users but is used only for internal flow control.


However, If we can get rid of `terminated` flag with it by introducing another state that express a state between `while (...)` and the head of finally block (such as `REQUESTING_TERMINATE`?) that would be the most cleaner (but would be a trade-off with simper states expression (for users use) vs more variants for internal state management).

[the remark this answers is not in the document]
ocadaruma: Hm, I see.
I think introducing another state to get rid of `terminated` has one more concern that it will violates this contract: https://github.com/line/decaton/blob/master/processor/src/main/java/com/linecorp/decaton/processor/SubscriptionStateListener.java#L83

Then introducing another flag as first suggestion sounds most appropriate for me.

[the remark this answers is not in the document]
ocadaruma: I think add an integration test to check expected state transition can cover this?
> preferred to have a test case proving that it won't happen

- partial invocation of rebalance listener never happen => i.e. `onChange(REBALANCING)` and `onChange(RUNNING)` always be invoked in this order in pair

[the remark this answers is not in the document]
kawamuray: > I think add an integration test to check expected state transition can cover this?

That's one good-to-have kind of test indeed.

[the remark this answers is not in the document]
kawamuray: >  I think introducing another state to get rid of terminated has one more concern that it will violates this contract: https://github.com/line/decaton/blob/master/processor/src/main/java/com/linecorp/decaton/processor/SubscriptionStateListener.java#L83

good point, I haven't came up with it.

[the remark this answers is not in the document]
ocadaruma: Added a test about state transition: https://github.com/line/decaton/pull/48/files#diff-ffc890431d9725fc4d648d1543fce5d0R38

processor/src/test/java/com/linecorp/decaton/processor/runtime/ProcessorSubscriptionTest.java
```diff
@@ -46,35 +42,21 @@
 
 public class ProcessorSubscriptionTest {
```
ocadaruma, line 43: While implementing state transition integration test, I noticed that `ConsumerRebalanceListener#onPartitionsRevoked` is no longer triggered when consumer startup. (i.e. transition to `RUNNING` from `INITIALIZING` directly).

- Fix also this test not to call `onPartitionsRevoked`
- Besides, I came up with it's better to simulate rebalance listener invocation by injecting pollTask in each test [like this](https://github.com/line/decaton/pull/48/files#diff-633ecc9f820b590d094c66ac75f62ef8R90) instead of trying to simulate by mock consumer impl

processor/src/test/java/com/linecorp/decaton/processor/runtime/ProcessorSubscriptionTest.java
```diff
@@ -101,28 +83,25 @@ private static ProcessorSubscription subscription(Consumer<String, byte[]> consu
...
     public void testStateTransition() throws Exception {
         TopicPartition tp = new TopicPartition("topic", 0);
-        Consumer<String, byte[]> consumer = spy(new DecatonMockConsumer());
+        DecatonMockConsumer consumer = new DecatonMockConsumer();
         List<State> states = Collections.synchronizedList(new ArrayList<>());
 
         CountDownLatch pollLatch = new CountDownLatch(1);
-        doAnswer(inv -> {
-            Object records = inv.callRealMethod();
+        consumer.schedulePollTask(() -> {
```
kawamuray, line 90: I see, such a function exists.

testing/src/main/java/com/linecorp/decaton/testing/TestUtils.java
```diff
@@ -137,14 +139,16 @@ private static Properties defaultProducerProps(String bootstrapServers) {
...
      * @param <T> type of tasks
      * @return {@link ProcessorSubscription} instance which is already running
      */
     public static <T> ProcessorSubscription subscription(String subscriptionId,
                                                          String bootstrapServers,
                                                          ProcessorsBuilder<T> processorsBuilder,
                                                          RetryConfig retryConfig,
-                                                         PropertySupplier propertySupplier) {
+                                                         PropertySupplier propertySupplier,
+                                                         SubscriptionStateListener stateListener) {
```
kawamuray, line 151: Umm.. feel like it start getting pointless and not usable to have this many of arguments in one method... 
We might have to change style passing builder to test code to let them do arbitrary thing in near future?

ocadaruma: Ya I'm thinking that too.

Cleaned up like this: https://github.com/line/decaton/pull/48/commits/4d269e8010873de93fc185ef2bfcb1567562ff7b

Issue #45, opened 2020-05-26
Introduce stable continuous benchmark and recover trackability of performance changes


kawamuray:
See: #44

kawamuray:
Should be considered done by https://line.github.io/decaton/

PR #49, opened 2020-06-23
Add manual git-based continuous benchmark

kawamuray:
This PR attempts to address #45 .

# Motivation
As I reverted the previous `PerformanceBoundTest` in https://github.com/line/decaton/pull/44, the way to run benchmark in CI showed a stability issue in terms of benchmark results.
It might be an option to pursue another CI service that provides stabler environment, but I think it's much better to take full control of benchmark environment that provides predictable resource and stable results.
To do that, I propose a new manual way of executing benchmark in our own servers along with the way to visualize trend of results that was also missing in our first attempt.

# Workflow
1. A commit pushed
2. A scheduled job in our own machine continuously polls for new commits
3. Once detected, job executes benchmark and store the result into a dedicated directory `commit-data/`
4. The job commits and pushes new file in `commit-data` into `gh-pages` branch.
5. If the result value regressed N% since the last run, it indicates us by leaving comment in PR or for a commit through github API (still in consideration).
6. We can access gh-page of this repo (to be created) to see the trend of performance transitions.

One of the downsides of this way is that we cannot detect regression while working on PR. However, since it's relatively more  difficult to properly judge performance regression is valid or not, I think it still make sense to deal with it not simply as "success/error" but in wider scope as trend.

This PR doesn't implements part 5 of the above steps. I'll follow it up later.
You can see demo of trend visualization at my personal fork:
- https://github.com/kawamuray/decaton/tree/gh-pages
- https://kawamuray.github.io/decaton/

### DIFF
--- a/cb/cb.sh
+++ b/cb/cb.sh
@@ -1,0 +1,92 @@
+#!/bin/bash
+set -eu
+
+config="${1:-}"
+if [ -z "$config" ]; then
+    echo "Usage: $0 PATH_TO_CONFIG" >&2
+    exit 1
+fi
+. "$config"
+
+base_dir=$(dirname $0)
+repo=$WORK_DIR/repo
+last_rev_file=$WORK_DIR/last_revision
+store_dir=$repo/$STORE_DIR
+
+function log() {
+    echo "[$(date +%Y-%m-%dT%H:%M:%S)] $*" >&2
+}
+
+function checkout() {
+    log "Checking out '$1'"
+    cleanup
+    git checkout "$1"
+}
+
+function pull() {
+    log "Pulling latest revision"
+    git pull
+}
+
+function load_last_revision() {
+    if [ ! -e $last_rev_file ]; then
+        log "Last revision file $last_rev_file missing, creating with current HEAD"
+        git rev-parse HEAD > $last_rev_file
+    fi
+    cat $last_rev_file
+}
+
+function revisions_since() {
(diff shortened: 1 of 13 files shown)


cb/cb.sh
```diff
@@ -0,0 +1,97 @@
...
+}
+
+function revisions_since() {
+    git rev-list --reverse $1..HEAD
+}
+
+function commit_rev() {
+    rev=$1
+    log "Committing revision $rev"
+    git --git-dir=$repo/.git --work-tree=$repo add $store_dir/$rev
```
ocadaruma, line 46: [nits] `git -C` more simpler?

kawamuray: Ah nice, point, sounds it works.

kawamuray:
> I wonder if it's appropriate to include "git-based publishing workflow" in Decaton repository
>    I think only Decaton maintainer executes this workflow, so not necessary to include in repo?

Not "necessarily" to put in here, but is just "fine" to?
* Users may wonder how benchmarks (showed in the performance page) are executed and they just can see it there
* We don't have to prepare a different place (separate repo?) to store just few such files
* Scripts may evolve along with decaton codebase, making it possible just to checkout the target revision to obtain a benchmark script that is compatible with that version of decaton codebase

while I don't see any cons to keep them.
Besides, I think it's common for OSS projects repo to store such "admin-only" stuff like release scripts.

PR #53, opened 2020-07-03
Commits offsets within consume-loop asynchronously to avoid blocking loop

kawamuray:
# Motivation

Decaton has been using `commitSync()` in pool-loop but there was some cases where it becomes a bottleneck especially when consumer's consuming in batch style where processing itself doesn't take much long due to latency awaiting OffsetCommit response from broker.

I think there's no point to wait offset commit's result synchronously in poll-loop except at 1. shutdown and 2. before rebalance.
This PR attempts to leverage async commit when decaton instance is consuming in regular loop so it won't be affected by high OffsetCommit latency from broker anymore.
There should be no essential difference with current way in terms of failure handling since we're anyway crushing exception from `commitSync()` now.

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
@@ -62,6 +62,7 @@ public class ProcessorSubscription extends Thread implements AsyncShutdownable {
     private final Property<Long> rebalanceTimeoutMillis;
     private final SubscriptionStateListener stateListener;
     private final SubscriptionMetrics metrics;
+    private volatile boolean asyncCommitInFlight;
 
     public ProcessorSubscription(SubscriptionScope scope,
                                  Supplier<Consumer<String, byte[]>> consumerSupplier,
@@ -111,14 +112,53 @@ public class ProcessorSubscription extends Thread implements AsyncShutdownable {
         log.debug("Committing offsets {}: {}", sync ? "SYNC" : "ASYNC", commitOffsets);
         if (sync) {
             consumer.commitSync(commitOffsets);
+            contexts.updateCommittedOffsets(commitOffsets);
         } else {
+            if (asyncCommitInFlight) {
+                // We would end up making multiple OffsetCommit requests containing same set of offsets if we proceed
+                // another async commit without waiting the first one to complete.
+                // This would be harmless if `decaton.commit.interval.ms` is set to sufficiently large, but otherwise
+                // we would make abusive offset commits despite there's no progress in processing.
+                log.debug("Skipping commit due to another async commit is currently in-flight");
+                return;
+            }
             consumer.commitAsync(commitOffsets, (offsets, exception) -> {
+                asyncCommitInFlight = false;
                 if (exception != null) {
                     log.warn("Offset commit failed asynchronously", exception);
+                    return;
+                }
+                // Exception raised from the commitAsync callback bubbles up to Consumer.poll() so it kills this
+                // subscription thread.
+                // Basically the exception thrown from here is considered a bug, but failing to commit offset itself
+                // isn't usually critical for consumer's continuation availability so we grab it here widely.
+                try {
+                    // Some thread and timing safety consideration about below operation.
+                    // Thread safety:
+                    // Below operation touches some thread-unsafe resources such as ProcessingContexts map and
+                    // variable storing committed offset in ProcessingContext but we can assume this callback is
+                    // thread safe because in all cases a callback for commitAsync is called from inside of
+                    // Consumer.poll() which is called only by this (subscription) thread.
+                    // Timing safety:
+                    // At the time this callback is triggered there might be a change in ProcessingContexts map
+                    // underlying contexts. It would occur in two cases:
+                    // 1. When group rebalance is triggered and the context is dropped by revoke at partitionsRevoked().
+                    // 2. When dynamic processor reload is triggered and the context is renewed by
+                    // PartitionContexts.maybeHandlePropertyReload.
+                    // The case 2 is safe (safe to keep updating committed offset in renewed PartitionContext) because
+                    // it caries previously consuming offset without reset.
+                    // The case 1 is somewhat suspicious but should still be safe, because whenever partition revoke
+                    // happens it calls commitSync() through onPartitionsRevoked(). According to the document of
+                    // commitAsync(), it is guaranteed that its callback is called before subsequent commitSync()
+                    // returns. So when a context is dropped from PartitionContexts, there should be no in-flight
+                    // commitAsync().
+                    contexts.updateCommittedOffsets(commitOffsets);
+                } catch (RuntimeException e) {
+                    log.error("Unexpected exception caught while updating committed offset", e);
                 }
             });
+            asyncCommitInFlight = true;
         }
-        contexts.updateCommittedOffsets(commitOffsets);
     }
 
     private void waitForRemainingTasksCompletion(long timeoutMillis) {
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
@@ -191,14 +191,15 @@ public class ProcessorSubscription extends Thread implements AsyncShutdownable {
             Duration elapsed = timer.duration();
             if (pendingTasksCount == 0) {
                 if (log.isDebugEnabled()) {
-                    log.debug("Waiting for task completion is successful {} ms", Utils.formatNum(elapsed.toMillis()));
+                    log.debug("Waiting for task completion is successful {} ms",
+                              Utils.formatNum(elapsed.toMillis()));
                 }
                 break;
             }
 
             if (elapsed.toNanos() >= timeoutNanos) {
                 log.warn("Timed out waiting {} tasks to complete after {} ms",
-                        Utils.formatNum(elapsed.toMillis()), pendingTasksCount);
+                         pendingTasksCount, Utils.formatNum(elapsed.toMillis()));
                 break;
             }
 
@@ -221,7 +222,7 @@ public class ProcessorSubscription extends Thread implements AsyncShutdownable {
         contexts.dropContexts(partitions);
         if (log.isInfoEnabled()) {
             log.info("Processed revoke {} partitions in {} ms",
-                    partitions.size(), Utils.formatNum(timer.elapsedMillis()));
+                     partitions.size(), Utils.formatNum(timer.elapsedMillis()));
         }
     }
 
@@ -236,7 +237,7 @@ public class ProcessorSubscription extends Thread implements AsyncShutdownable {
         }
         if (log.isInfoEnabled()) {
             log.info("Processed assign {} partitions in {} ms",
-                    partitions.size(), Utils.formatNum(timer.elapsedMillis()));
+                     partitions.size(), Utils.formatNum(timer.elapsedMillis()));
         }
     }
 
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
@@ -62,7 +62,7 @@ public class ProcessorSubscription extends Thread implements AsyncShutdownable {
     private final Property<Long> rebalanceTimeoutMillis;
     private final SubscriptionStateListener stateListener;
     private final SubscriptionMetrics metrics;
-    private volatile boolean asyncCommitInFlight;
+    private boolean asyncCommitInFlight;
 
     ProcessorSubscription(SubscriptionScope scope,
                           Supplier<Consumer<String, byte[]>> consumerSupplier,
@@ -139,6 +139,14 @@ public class ProcessorSubscription extends Thread implements AsyncShutdownable {
                     log.warn("Offset commit failed asynchronously", exception);
                     return;
                 }
+                if (Thread.currentThread() != this) {
+                    // This isn't expected to happen (see below comment) but we check it with cheap cost
+                    // just in case to avoid silent corruption.
+                    log.error("BUG: commitAsync callback triggered by an unexpected thread: {}." +
+                              " Please report this to Decaton developers", Thread.currentThread());
+                    return;
+                }
+
                 // Exception raised from the commitAsync callback bubbles up to Consumer.poll() so it kills this
                 // subscription thread.
                 // Basically the exception thrown from here is considered a bug, but failing to commit offset itself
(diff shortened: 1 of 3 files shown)


processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -103,14 +101,22 @@ private void updateState(SubscriptionStateListener.State newState) {
...
         }
-        logger.debug("Committing offsets: {}", commitOffsets);
-        consumer.commitSync(commitOffsets);
+        log.debug("Committing offsets {}: {}", sync ? "SYNC" : "ASYNC", commitOffsets);
+        if (sync) {
+            consumer.commitSync(commitOffsets);
+        } else {
+            consumer.commitAsync(commitOffsets, (offsets, exception) -> {
+                if (exception != null) {
+                    log.warn("Offset commit failed asynchronously", exception);
```
mauhiz, line 116: is it worth logging the offsets?

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -103,14 +101,22 @@ private void updateState(SubscriptionStateListener.State newState) {
...
         Map<TopicPartition, OffsetAndMetadata> commitOffsets = contexts.commitOffsets();
         if (commitOffsets.isEmpty()) {
-            logger.debug("No new offsets to commit, skipping commit");
+            log.debug("No new offsets to commit, skipping commit");
             return;
         }
-        logger.debug("Committing offsets: {}", commitOffsets);
-        consumer.commitSync(commitOffsets);
+        log.debug("Committing offsets {}: {}", sync ? "SYNC" : "ASYNC", commitOffsets);
+        if (sync) {
```
mauhiz, line 111: why do we keep the sync way? (does it have any merit?)

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
kawamuray: `offsets.toString()` is really a debug string with unpredictable length (proportional to number of partitions) and I don't think it's good idea to put it in level >= warn which is likely enabled in most of applications.
Also,
* Even if we get offset values that are failed to commit, I don't think there's almost nothing we can do with it
* With much lower logging level (debug) it's written once in above debug logging

any thoughts?

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
kawamuray: Yes. Currently we call offset commit at following timings and we need below 2 to be synchronous to avoid unnecessary duplication (re-consume) of records because for those cases basically it's the last chance to checkpoint the completed offsets.
* Every 1 second in consumer.poll() => queue loop
* At group rebalance (involves partition reassignment)
* At shutdown

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
mauhiz: I see. I guess we can get enough clues from the exception itself!

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
mauhiz: I see. I didn't try by myself so it's pure speculation but as you postulated for Shutdown, I'd have expected that rebalancing would wait for pending commit requests to be completed before actually proceeding.
Anyways, both cases are pretty rare so keeping sync will not affect performance. It's just about cleaning up.

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -103,14 +101,22 @@ private void updateState(SubscriptionStateListener.State newState) {
...
+            consumer.commitSync(commitOffsets);
+        } else {
+            consumer.commitAsync(commitOffsets, (offsets, exception) -> {
+                if (exception != null) {
+                    log.warn("Offset commit failed asynchronously", exception);
+                }
+            });
+        }
         contexts.updateCommittedOffsets(commitOffsets);
```
mauhiz, line 120: I think I vaguely understand, but can you comment as to why this is synchronous and not a callback of `commitAsync` ?

kawamuray: Oh, I think this should be handled differently for async case (do in callback). Let me confirm and fix, thanks for nice catch!

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -103,15 +113,63 @@ private void updateState(SubscriptionStateListener.State newState) {
...
+                    // happens it calls commitSync() through onPartitionsRevoked(). According to the document of
+                    // commitAsync(), it is guaranteed that its callback is called before subsequent commitSync()
+                    // returns. So when a context is dropped from PartitionContexts, there should be no in-flight
+                    // commitAsync().
+                    contexts.updateCommittedOffsets(commitOffsets);
+                } catch (RuntimeException e) {
+                    log.error("Unexpected exception caught while updating committed offset", e);
+                }
+            });
+            asyncCommitInFlight = true;
```
mauhiz, line 171: If you are unlucky with thread scheduling, there is change that this is reached after the callback is executed. I think that you cannot use a mere volatile field and need AtomicBoolean with CAS

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -103,15 +113,63 @@ private void updateState(SubscriptionStateListener.State newState) {
...
+                    // 2. When dynamic processor reload is triggered and the context is renewed by
+                    // PartitionContexts.maybeHandlePropertyReload.
+                    // The case 2 is safe (safe to keep updating committed offset in renewed PartitionContext) because
+                    // it caries previously consuming offset without reset.
+                    // The case 1 is somewhat suspicious but should still be safe, because whenever partition revoke
+                    // happens it calls commitSync() through onPartitionsRevoked(). According to the document of
+                    // commitAsync(), it is guaranteed that its callback is called before subsequent commitSync()
+                    // returns. So when a context is dropped from PartitionContexts, there should be no in-flight
+                    // commitAsync().
+                    contexts.updateCommittedOffsets(commitOffsets);
```
mauhiz, line 166: since this now async, different threads can call this. I suppose the actual `PartitionContext.lastCommittedOffset` should now be volatile?

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
kawamuray: That shouldn't happen too. Also explained in the comment https://github.com/line/decaton/pull/53/files#diff-bd5dd2f1469e6022198c76787fd059eaR147

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
kawamuray: No it can't. Please read through this comment explaining what I observed from reading kafka code: https://github.com/line/decaton/pull/53/files#diff-bd5dd2f1469e6022198c76787fd059eaR147

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -103,15 +113,63 @@ private void updateState(SubscriptionStateListener.State newState) {
...
+                asyncCommitInFlight = false;
+                if (exception != null) {
+                    log.warn("Offset commit failed asynchronously", exception);
+                    return;
+                }
+                // Exception raised from the commitAsync callback bubbles up to Consumer.poll() so it kills this
+                // subscription thread.
+                // Basically the exception thrown from here is considered a bug, but failing to commit offset itself
+                // isn't usually critical for consumer's continuation availability so we grab it here widely.
+                try {
```
ocadaruma, line 146: Hm, this try-catch block changes error handling semantics a bit from current impl.

current: `commitOffsetsIfNecessary` catches only `CommitFailedException | TimeoutException` so unexpected runtime exception from `contexts .updateCommittedOffsets` would kill subscription thread

And I'm not sure if it's good to catch widely here, since exception from updateCommittedOffset is completely unexpected (likely bug as you mentioned).

`updateCommittedOffset` just simply iterating over passed map and update each context's state, so exception from this method would be a sign of contexts state corruption which might be more critical?

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
mauhiz: Believe it or not I did not notice your novel over here. Sorry about that..
In any case, I would feel safer with an actual test that checks the thread in the callback; although it sounds unlikely that Kafka client would change its threading model?

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
mauhiz: 👍

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
mauhiz: It is a case of acute TLDRitis, I'm afraid 😭

kawamuray: >  although it sounds unlikely that Kafka client would change its threading model?

KafkaConsumer itself is basically "for a single thread" component. It doesn't have background IO thread (except a thread dedicated for heartbeat), and from what we can see in implementation of commitAsync callback it is obviously controlled it's triggering timing and location with intention:
* https://github.com/apache/kafka/blob/6f3ceea5c61d90d4075ed5c5a22bde33142f7ec7/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerCoordinator.java#L1452 (callback triggered here)
* https://github.com/apache/kafka/blob/6f3ceea5c61d90d4075ed5c5a22bde33142f7ec7/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerCoordinator.java#L900 (the entrypoint calling .invoke())
* `invokeCompletedOffsetCommitCallbacks` is called only by `poll`, `commitXX` and `close`

Reading this implementation I think it will be unlikely that they change implementation in the future without major, noticable breaking change.

> In any case, I would feel safer with an actual test that checks the thread in the callback;

Well it's not that easy to add a test that confirms it but I just came up with one cheap assert that we can log when it happens at least. f748ead

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
kawamuray: Good point. The only reason that I did this is because I thought likelihood of hitting bug is very different between sync (kind of procedural) and async case.
As we know the whole implementation of `updateCommittedOffsets` in case of sync we would almost never expect an exception to be raised, so that to me, it's kind of "really unexpected and have no idea what's going on" (so we should stop entire execution immediately).

> And I'm not sure if it's good to catch widely here, since exception from updateCommittedOffset is completely unexpected (likely bug as you mentioned).

The reason I'm trying to preserve execution even when we meet an unexpected exception here is explained in the above comment.

> updateCommittedOffset just simply iterating over passed map and update each context's state, so exception from this method would be a sign of contexts state corruption which might be more critical?

Exactly. However I think it still has some possibility to recover at next or few times later attempt because an exception possibly we see here is by accessing some missing element in contexts map that might happen possibly (by a bug) by timing issue rather than a permanent one. (e.g, missing context implies on next commit attempt the partition's offset won't be the subject of commit)

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
mauhiz: Thanks. 👍

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -124,13 +190,13 @@ private void waitForRemainingTasksCompletion(long timeoutMillis) {
             final int pendingTasksCount = contexts.totalPendingTasks();
             Duration elapsed = timer.duration();
             if (pendingTasksCount == 0) {
-                logger.debug("waiting for task completion is successful {} ns \\(^^)/",
-                             Utils.formatNanos(elapsed));
+                log.debug("waiting for task completion is successful {} ns \\(^^)/",
```
ocadaruma, line 193: [nits] Isn't it time to get rid of this ascii art?

processor/src/test/java/com/linecorp/decaton/processor/runtime/ProcessorSubscriptionTest.java
```diff
@@ -106,4 +136,148 @@ public void testStateTransition() throws Exception {
...
+        SubscriptionScope scope = scope("topic");
+        ProcessorSubscription subscription = new ProcessorSubscription(
+                scope,
+                () -> consumerMock,
+                null,
+                scope.props(),
+                null,
+                contextsMock) {
+            @Override
+            public void run() {
```
ocadaruma, line 150: I understand this, but overriding `run` for test (though `run` contains important logic actually) looks like sign that we should refactor ProcessorSubscription to separate its responsibility into few components, for better testability...

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
kawamuray: lol okay.

processor/src/test/java/com/linecorp/decaton/processor/runtime/ProcessorSubscriptionTest.java
kawamuray: Agree. Do you want me to try that refactoring in this PR or leave it for later effort?

ocadaruma: Hm, I think it will make big diff, and we shouldn't include 2 significant changes about ProcessorSubscription in 1 PR

So let's do later

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
kawamuray: a9e15c9 with some extra refactoring :)

PR #59, opened 2020-07-31
Prevent NPE when closing Metrics twice

### LINKED ISSUE #58: Closing ProcessorSubscription twice causes NPE
trace:

```
java.lang.NullPointerException: null
	at com.linecorp.decaton.processor.metrics.Metrics$AbstractMetrics.close(Metrics.java:84)
	at com.linecorp.decaton.processor.metrics.Metrics$SubscriptionMetrics.close(Metrics.java:95)
	at com.linecorp.decaton.processor.runtime.ProcessorSubscription.awaitShutdown(ProcessorSubscription.java:424)
	at com.linecorp.decaton.processor.runtime.AsyncShutdownable.close(AsyncShutdownable.java:27)
```

It looks like closed `meter` should be removed from `meters` List in `AbstractMetrics`.

mauhiz:
Fixes #58 

- Clear `meters` on `close()`
- Since `meters` is now mutable, synchronize `close()` and `meter()`
- Make `meterRefCounts` non-synchronized to avoid double locking
- Add nullcheck for counters in `meterRefCounts`

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
@@ -77,7 +77,7 @@ public class Metrics {
         public void close() {
             synchronized (meterRefCounts) {
                 // traverse from the end to avoid arrayCopy
-                for (ListIterator<Meter> iterator = meters.listIterator(meters.size() - 1); iterator.hasNext(); ) {
+                for (ListIterator<Meter> iterator = meters.listIterator(meters.size()); iterator.hasPrevious(); ) {
                     Meter meter = iterator.previous();
                     Id id = meter.getId();
                     AtomicInteger count = meterRefCounts.get(id);
--- a/processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
@@ -18,9 +18,9 @@ package com.linecorp.decaton.processor.metrics;
 
 import java.time.Duration;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 
@@ -59,36 +59,30 @@ public class Metrics {
      * the instance from {@link MeterRegistry} when the last reference to the meter closed and disappeared.
      */
     abstract static class AbstractMetrics implements AutoCloseable {
-        private static final Map<Meter.Id, AtomicInteger> meterRefCounts = new HashMap<>();
-        private final List<Meter> meters;
+        private static final ConcurrentMap<Id, AtomicInteger> meterRefCounts = new ConcurrentHashMap<>();
+        private final List<Meter> meters = new ArrayList<>();
 
-        protected AbstractMetrics() {
-            meters = new ArrayList<>();
-        }
-
-        <T extends Meter> T meter(Supplier<T> ctor) {
-            synchronized (meterRefCounts) {
-                T meter = ctor.get();
-                meterRefCounts.computeIfAbsent(meter.getId(), key -> new AtomicInteger())
-                              .incrementAndGet();
-                meters.add(meter);
-                return meter;
-            }
+        synchronized <T extends Meter> T meter(Supplier<T> ctor) {
+            T meter = ctor.get();
+            meterRefCounts.computeIfAbsent(meter.getId(), key -> new AtomicInteger())
+                          .incrementAndGet();
+            meters.add(meter);
+            return meter;
         }
 
         @Override
-        public void close() {
-            synchronized (meterRefCounts) {
-                for (Meter meter : meters) {
-                    Meter.Id id = meter.getId();
-                    int count = meterRefCounts.get(id).decrementAndGet();
-                    if (count == 0) {
-                        meterRefCounts.remove(id);
-                        registry.remove(meter);
-                        meter.close();
-                    }
+        public synchronized void close() {
+            for (Meter meter : meters) {
+                Id id = meter.getId();
+                AtomicInteger count = meterRefCounts.get(id);
+                if (count != null && count.decrementAndGet() <= 0) {
+                    meterRefCounts.remove(id);
+                    registry.remove(meter);
+                    meter.close();
                 }
             }
+            // make close idempotent
+            meters.clear();
         }
     }
 
--- a/processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
@@ -18,9 +18,10 @@ package com.linecorp.decaton.processor.metrics;
 
 import java.time.Duration;
 import java.util.ArrayList;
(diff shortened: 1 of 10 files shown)


kawamuray:
Thanks for the PR :) 

> Since meters is now mutable, synchronize close() and meter()

I think it was synchronized among them already in as-is impl by synchronize against `refCount`?

> Make meterRefCounts non-synchronized to avoid double locking

Does it mean as a consequence of adding `synchronize` at method level? (by keeping static ref lock in current impl)

processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
```diff
@@ -59,36 +59,30 @@ private Metrics(AvailableTags availableTags) {
...
-                              .incrementAndGet();
-                meters.add(meter);
-                return meter;
-            }
+        private static final ConcurrentMap<Id, AtomicInteger> meterRefCounts = new ConcurrentHashMap<>();
+        private final List<Meter> meters = new ArrayList<>();
+
+        synchronized <T extends Meter> T meter(Supplier<T> ctor) {
+            T meter = ctor.get();
+            meterRefCounts.computeIfAbsent(meter.getId(), key -> new AtomicInteger())
```
kawamuray, line 67: It would be (by current usage) not realistic to happen but this concurrency semantics feels bit broken because it would leaks a reference count in following situation?

A and B are operating on different `XXMetrics extends AbstractMetrics` instance.

1. Thread A ` AtomicInteger count = meterRefCounts.get(id); `
2. Thread B `meterRefCounts.computeIfAbsent(meter.getId(), key -> new AtomicInteger()) => returns existing`
3. Thread A `count.decrementAndGet() <= 0) && meterRefCounts.remove(id);`
4. Thread B ` .incrementAndGet();`

So I think it's better to keep `meterRefCounts.computeIfAbsent(meter.getId(), key -> new AtomicInteger()).incrementAndGet()` this whole operation atomic among all `XXMetrics` instances by keeping synchronize on `meterRefCounts`.

processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
```diff
@@ -59,36 +59,30 @@ private Metrics(AvailableTags availableTags) {
...
-                    if (count == 0) {
-                        meterRefCounts.remove(id);
-                        registry.remove(meter);
-                        meter.close();
-                    }
+        public synchronized void close() {
+            for (Meter meter : meters) {
+                Id id = meter.getId();
+                AtomicInteger count = meterRefCounts.get(id);
+                if (count != null && count.decrementAndGet() <= 0) {
```
kawamuray, line 78: I think actually getting `null` here implies partial failure or sort of race.
How about iterating meters over in reverse order (from tail to head) with `list.listIterator()` and remove tail elements step-by-step with `iter.remove()`? Then getting null here becomes a real bug that should not be crushed I think.

processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
mauhiz: Okay. I think we can use `ConcurrentMap.computeIfPresent` which is atomic.
I don't want to nest synchronization

processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
mauhiz: It's a close method and as long as it does the job we don't care too much?
Let me first see if we can get away with `computeIfPresent`.

processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
kawamuray: You can just get rid of method level synchronize in case? Why do we needed to replace `synchronized` on `refcount` with method level synchornize in the first place?

processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
mauhiz: It does not seem reliable as `ConcurrentMap.computeIfPresent` says it may run remapping function several times (although I don't think it holds for `ConcurrentHashMap.computeIfPresent`).
Let me follow your idea there

mauhiz: > Then getting null here becomes a real bug

I don't think so, races can still happen since this static map is shared by several `AbstractMetrics` instances

processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
mauhiz: Ah, I see what you mean. We can indeed synchronize only on this static `meterRefCounts`

processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
```diff
@@ -79,14 +76,21 @@ protected AbstractMetrics() {
         @Override
         public void close() {
             synchronized (meterRefCounts) {
-                for (Meter meter : meters) {
-                    Meter.Id id = meter.getId();
-                    int count = meterRefCounts.get(id).decrementAndGet();
-                    if (count == 0) {
+                // traverse from the end to avoid arrayCopy
+                for (ListIterator<Meter> iterator = meters.listIterator(meters.size() - 1); iterator.hasNext(); ) {
```
kawamuray, line 80: `iterator.hasPrevious`?

kawamuray: and starting index should be `meters.size()`?

image

processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
```diff
@@ -79,14 +76,21 @@ protected AbstractMetrics() {
...
+                    Meter meter = iterator.previous();
+                    Id id = meter.getId();
+                    AtomicInteger count = meterRefCounts.get(id);
+                    if (count == null) {
+                        throw new IllegalStateException("Missing reference to meter: " + id);
```
kawamuray, line 85: > I don't think so, races can still happen since this static map is shared by several AbstractMetrics instances

hm, can it?
I thought current synchronize makes sure atomic 1. refcount decrement, 2. `meterRefCounts` entry drop and 3. `iterator.remove();` without possibility of partial failure in the middle so we never expect to see null here.

kawamuray:
btw the build is failing due to our outdated docs detection system - it fails the build if there's any update in dependent module since last confirmed version: `:base_version: 0.0.39` https://raw.githubusercontent.com/line/decaton/master/docs/getting-started.adoc

Please replace `:base_version: 0.0.39`  this with `:base_version: 0.0.40-SNAPSHOT` so the build passes.

processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
mauhiz: yes to both..

processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
mauhiz: Yes, my comment is deprecated. Sorry.
Hence throwing here meaning "wtf it's impossible"

PR #61, opened 2020-08-05
Make key deserializer of KafkaConsumer configurable

techno:
Hello, decaton team.

We are consuming non-decaton topic using decaton processor.
https://github.com/line/decaton/blob/master/docs/consuming-any-data.adoc

Decaton only accepts `String` key for consuming topic now.
Current implementation is using hard-coded `StringDeserializer`.
However, Our topic is using `Long` type key.
We want to use `ProcessingContext.key()` inside the processor, but it will be broken string.

This patch enables to handle any type of the key correctly.
You can specify original `Deserializer` to `KafkaConsumer` by consumer property `key.deserializer`.
This patch will not change consumer key type from `String` for compatibility.
So `key.deserializer` property only accepts subclass of `Deserializer<String>`.

In our case, we can use a arbitrary deserializer such as `LongToStringDeserializer` for `Long` key topics by this patch.
It's better than broken string.
This is just a special use case of decaton, so I don't think decaton should handle the key by raw type for non-decaton topics.

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/ProcessorsBuilder.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/ProcessorsBuilder.java
@@ -27,6 +27,7 @@ import com.linecorp.decaton.processor.runtime.DefaultTaskExtractor;
 import com.linecorp.decaton.processor.runtime.ProcessorScope;
 import com.linecorp.decaton.processor.runtime.ProcessorSubscription;
 import com.linecorp.decaton.processor.runtime.Processors;
+import com.linecorp.decaton.processor.serialization.StringDeserializer;
 import com.linecorp.decaton.processor.runtime.SubscriptionBuilder;
 
 import lombok.Getter;
@@ -41,13 +42,18 @@ import lombok.experimental.Accessors;
 public class ProcessorsBuilder<T> {
     @Getter
     private final String topic;
+    private final Deserializer<String> keyDeserializer;
     private final TaskExtractor<T> taskExtractor;
     private final TaskExtractor<T> retryTaskExtractor;
 
     private final List<DecatonProcessorSupplier<T>> suppliers;
 
-    public ProcessorsBuilder(String topic, TaskExtractor<T> taskExtractor, TaskExtractor<T> retryTaskExtractor) {
+    public ProcessorsBuilder(String topic,
+                             Deserializer<String> keyDeserializer,
+                             TaskExtractor<T> taskExtractor,
+                             TaskExtractor<T> retryTaskExtractor) {
         this.topic = topic;
+        this.keyDeserializer = keyDeserializer;
         this.taskExtractor = taskExtractor;
         this.retryTaskExtractor = retryTaskExtractor;
         suppliers = new ArrayList<>();
@@ -63,7 +69,7 @@ public class ProcessorsBuilder<T> {
      */
     public static <T> ProcessorsBuilder<T> consuming(String topic, Deserializer<T> deserializer) {
         DefaultTaskExtractor<T> taskExtractor = new DefaultTaskExtractor<>(deserializer);
-        return new ProcessorsBuilder<>(topic, taskExtractor, taskExtractor);
+        return new ProcessorsBuilder<>(topic, new StringDeserializer(), taskExtractor, taskExtractor);
     }
 
     /**
@@ -75,6 +81,21 @@ public class ProcessorsBuilder<T> {
      * @return an instance of {@link ProcessorsBuilder}.
      */
     public static <T> ProcessorsBuilder<T> consuming(String topic, TaskExtractor<T> taskExtractor) {
+        return consuming(topic, new StringDeserializer(), taskExtractor);
+    }
+
+    /**
+     * Create new {@link ProcessorsBuilder} that consumes message from topic expecting tasks of type
+     * which can be parsed by valueParser.
+     * @param topic the name of topic to consume.
+     * @param keyDeserializer the deserializer for key that implements {@link Deserializer}.
+     * @param taskExtractor the extractor to extract task of type {@link T} from message bytes.
+     * @param <T> the type of instantiated tasks.
+     * @return an instance of {@link ProcessorsBuilder}.
+     */
+    public static <T> ProcessorsBuilder<T> consuming(String topic,
+                                                     Deserializer<String> keyDeserializer,
+                                                     TaskExtractor<T> taskExtractor) {
         DefaultTaskExtractor<byte[]> innerExtractor = new DefaultTaskExtractor<>(bytes -> bytes);
         TaskExtractor<T> retryTaskExtractor = bytes -> {
             // Retry tasks are serialized as PB DecatonTaskRequest.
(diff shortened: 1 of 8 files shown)


kawamuray:
Thanks for the PR!

Non-string key is definitely a thing that decaton should support. The current situation - hardcoded String keys - is a heritage from the early days of decaton that it didn't supported arbitrary topic consumption and has putting assumption at reach record's data format both for key and value.
As of today, decaton officially supports arbitrary topic consumption like a general-purpose consumer framework so it should also support arbitrary format key as well. Given that, 

>  This patch will not change consumer key type from String for compatibility.

I don't think we need to try keeping this. However changing key as generic type `T` consequence to mess almost all classes of decaton to be a generic class which we would wanna avoid.
Changing key to be `byte[]` would be a good option. Some downsides are:
- Broad amount of code needs to be changed
- Some debug logging becomes less-useful by losing human readable key output

I think all existing logic that depends on key like hashing, key lookup still works with `byte[]` but not 100% sure yet.

If you're motivated to proceed that we can discuss deeper design before going further.
If you prefer to quickly workaround for your use I can review based on current changers too.
WDYT?

processor/src/main/java/com/linecorp/decaton/processor/runtime/ConsumerSupplier.java
```diff
@@ -44,11 +47,14 @@ public ConsumerSupplier(Properties config) {
 
     @Override
     public Consumer<String, byte[]> get() {
-        return new KafkaConsumer<>(mergedProps(), new StringDeserializer(), new ByteArrayDeserializer());
+        return new KafkaConsumer<>(mergedProps(), null, new ByteArrayDeserializer());
```
kawamuray, line 50: This is bit tricky and dangerous bypass of type check as in we cannot ensure the deserializer specified in properties actually returns `String`.
If it wasn't, I'm not sure at which place does the first error gets thrown, that might makes users confused what has went wrong.
Would it better to take key deserializer (`Function<byte[], String>`) at `ProcessorBuilder` explicitly?

techno: I see, added a parameter `keyDeserializer` to `ProcessorsBuilder.consuming()`

techno:
Updated changes to consume with `Consumer<bytes[], bytes[]>` from topic.
However, this changes is still using `String` for key.
So keys will be deserialized to `String` by the deserializer specified in `ProcessorsBuilder.consuming()`.

> I think all existing logic that depends on key like hashing, key lookup still works with byte[] but not 100% sure yet.

It seems difficult to implement `blacklistedKeysFilter` for `bytes[]`.
Do you have any ideas?

kawamuray:
> It seems difficult to implement blacklistedKeysFilter for bytes[].

I think we have 2 possible options.
1. Let the dynamic property driver (centraldogma) to convert `String` into `bytes[]`. Comparison of keys should still work as bytes.
2. Support key blacklisting only if the key is the type of `String` (the filter just tries to instantiate `String` from whatever the byte array is and do comparison which might always fail when key is not a type of String).

However, given that the current mechanism used for obtaining keys list is `Property` which obviously intends to deal with human-readaible/editable properties I think the 2nd option would be more suitable here, if we do that.

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -344,20 +344,21 @@ public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
...
-        ConsumerRecords<String, byte[]> records = consumer.poll(POLL_TIMEOUT_MILLIS);
+        ConsumerRecords<byte[], byte[]> records = consumer.poll(POLL_TIMEOUT_MILLIS);
         metrics.consumerPollTime.record(timer.duration());
 
         timer = Utils.timer();
         records.forEach(record -> {
             TopicPartition tp = new TopicPartition(record.topic(), record.partition());
             PartitionContext context = contexts.get(tp);
             DeferredCompletion completion = context.registerOffset(record.offset());
+            String key = processors.keyDeserializer().deserialize(record.key());
```
kawamuray, line 357: `Processors` contains stuff that becomes necessary only at just before `DecatonProcessor` begins to process tasks. That is, other parent parts of decaton stays agnostic about what it does for the given task which is yet represented as `byte[]`. 
In contract, the key in the current impl is expected to be read as `String` in not only the parts around  `DecatonProcessor` but also at nearly everywhere of decaton impl starting from `ProcessorSubscription` to downwards.
So I think `keyDeserializer` should be a properly of `ProcessorSubscription` itself rather than one of `Processors` (hence `ProcessorBuilder` either).

processor/src/main/java/com/linecorp/decaton/processor/ProcessorsBuilder.java
```diff
@@ -75,6 +81,21 @@ public ProcessorsBuilder(String topic, TaskExtractor<T> taskExtractor, TaskExtra
...
      */
     public static <T> ProcessorsBuilder<T> consuming(String topic, TaskExtractor<T> taskExtractor) {
+        return consuming(topic, new StringDeserializer(), taskExtractor);
+    }
+
+    /**
+     * Create new {@link ProcessorsBuilder} that consumes message from topic expecting tasks of type
+     * which can be parsed by valueParser.
+     * @param topic the name of topic to consume.
+     * @param keyDeserializer the deserializer for key that implements {@link Deserializer}.
```
kawamuray, line 91: Let's mention that whatever the representation of key in the record is, it needs to be instantiated into a `String`.

processor/src/main/java/com/linecorp/decaton/processor/runtime/Processors.java
```diff
@@ -33,15 +34,18 @@
 
     private final List<DecatonProcessorSupplier<T>> suppliers;
     private final DecatonProcessorSupplier<byte[]> retryProcessorSupplier;
+    private final Deserializer<String> keyDeserializer;
```
kawamuray, line 37: Mentioned in the above comment as well but I think this should be a property of `ProcessorSubscription`.

processor/src/main/java/com/linecorp/decaton/processor/serialization/StringDeserializer.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package com.linecorp.decaton.processor.serialization;
+
+import com.linecorp.decaton.common.Deserializer;
+
+public class StringDeserializer implements Deserializer<String> {
+    @Override
+    public String deserialize(byte[] bytes) {
+        if (bytes == null)
```
kawamuray, line 24: In our (LINE's standard though :p) coding style always requires brackets.

processor/src/main/java/com/linecorp/decaton/processor/serialization/StringDeserializer.java
```diff
@@ -0,0 +1,29 @@
...
+            return null;
+        else
+            return new String(bytes);
```
kawamuray, line 27: `new String(bytes, StandardCharsets.UTF_8)` to be bit more conservative?

kawamuray:
> Changing key to be byte[] would be a good option. 

On my 2nd thought looking at your current impl holding internal key type as `String` while letting to deserialize arbitrary thing into it would be a well-balanced option as well.

ocadaruma:
@techno Thanks for the PR.

I guess the original motivation of this PR is covered by https://github.com/line/decaton/pull/157 so we can close this PR for now.

Please feel free to ask us or raise another issue if necessary.

PR #65, opened 2020-09-10
Publish decaton testing

ocadaruma:
## Summary
- `testing` module is useful not only for decaton-internal test, but also general Kafka-client integration testing
- publish `testing` module to make writing integration-test for users


ocadaruma:
I don't have a plan to proceed this now. Let me close once

PR #68, opened 2020-09-29
Publish benchmark module locally to add capability to implement custom runner

ocadaruma:
## Motivation
- `benchmark` module is designed to be able to run arbitrary `Runner` implementation, not only built-in `DecatonRunner`
- But currently, running custom Runner implementation cannot be done without modifying decaton repository itself since `benchmark` module is not published
- This PR publishes `benchmark` module locally to be able to depend on it from custom runner

### publish `benchmark` module only locally
- not to maven central. to avoid a lot of redistributions
### communicate ForkingExecution <-> benchmarker via RMI
- change interprocess communication to RMI
- currently, interprocess communication is done through stdout
  * but this forces custom runner not to output any information to stdout (and it isn't documented anywhere)
  * we shouldn't restrict use of stdout since we usually output debug information to stdout

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -13,11 +13,105 @@ Build a shadowJar:
 
 Run with arbitrary parameters:
 ```sh
-./debm.sh
+./debm.sh \
  --title "Decaton" \
  --runner com.linecorp.decaton.benchmark.DecatonRunner \
  --tasks 10000 \
  --warmup 100 \
  --simulate-latency=10 \
  --param=decaton.partition.concurrency=20 
+```
+
+How to run benchmark with custom Runner
+=======================================
+
+#### Publish `benchmark` module locally:
+
+```sh
+cd /path/to/decaton/benchmark
+
+../gradlew :benchmark:publishToMavenLocal -x sign
+```
+
+#### Add `decaton-benchmark` dependency to your custom runner project:
+
+build.gradle
+
+```groovy
+repositories {
+    mavenCentral()
+    mavenLocal()
+}
+
+dependencies {
+    implementation "com.linecorp.decaton:decaton-benchmark:$DECATON_VERSION"
+}
+```
+
+#### Implement com.linecorp.decaton.benchmark.Runner:
+
+CustomRunner.java
+
+```java
+package com.example;
+
+import com.linecorp.decaton.benchmark.Recording;
+import com.linecorp.decaton.benchmark.ResourceTracker;
+import com.linecorp.decaton.benchmark.Runner;
+import com.linecorp.decaton.benchmark.Task;
+
+public class CustomRunner implements Runner {
+    @Override
+    public void init(Config config, Recording recording, ResourceTracker resourceTracker)
+            throws InterruptedException {
+        // initialize and start your consumer
+    }
+
+    @Override
+    public void close() throws Exception {
+        // cleanup
+    }
+}
+```
+
+#### Build a shadowJar of the custom runner:
+
+build.gradle
+
+```groovy
+buildscript {
+    repositories {
+        jcenter()
+    }
+    dependencies {
+        classpath 'com.github.jengelman.gradle.plugins:shadow:5.1.0'
+    }
+}
+
+apply plugin: 'com.github.johnrengelman.shadow'
+```
+
+```sh
+cd /path/to/custom-runner
+
+./gradlew shadowJar
+```
+
+#### Build a shadowJar of the benchmark project:
+```sh
+cd /path/to/decaton/benchmark
+
+../gradlew :benchmark:shadowJar
+```
+
+Run:
+```sh
+export CLASSPATH=/path/to/custom-runner-all.jar
+
+./debm.sh \
+ --title "Custom" \
+ --runner com.example.CustomRunner \
+ --tasks 10000 \
+ --warmup 100 \
+ --simulate-latency=10 
 ```
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -97,16 +97,12 @@ cd /path/to/custom-runner
 ./gradlew shadowJar
 ```
 
-#### Build a shadowJar of the benchmark project:
-```sh
-cd /path/to/decaton/benchmark
+### Run benchmark:
 
-../gradlew :benchmark:shadowJar
-```
-
-Run:
 ```sh
 export CLASSPATH=/path/to/custom-runner-all.jar
+
+cd /path/to/decaton/benchmark
 
 ./debm.sh \
  --title "Custom" \
--- a/benchmark/src/main/java/com/linecorp/decaton/benchmark/BenchmarkRmi.java
+++ b/benchmark/src/main/java/com/linecorp/decaton/benchmark/BenchmarkRmi.java
@@ -32,6 +32,8 @@ import java.util.concurrent.LinkedBlockingQueue;
 
 import com.linecorp.decaton.benchmark.Execution.Stage;
 
+import lombok.Getter;
+import lombok.experimental.Accessors;
 import lombok.extern.slf4j.Slf4j;
 
 /**
@@ -45,8 +47,13 @@ class BenchmarkRmi implements AutoCloseable {
     private static final String REMOTE_NAME = "BenchmarkRmi";
 
     private final BlockingQueue<Stage> stageQueue = new LinkedBlockingQueue<>();
-    private int port;
     private RemoteCallback callback;
+    /**
+     * Port of the RMI registry
+     */
+    @Getter
+    @Accessors(fluent = true)
+    private int port;
 
     interface RemoteCallback extends Remote {
         /**
@@ -74,14 +81,6 @@ class BenchmarkRmi implements AutoCloseable {
             this.port = sock.getLocalPort();
             return sock;
         }
-    }
-
-    /**
-     * Port of the RMI registry
-     * @return port
-     */
-    int port() {
-        return port;
     }
 
     /**
@@ -158,7 +157,7 @@ class BenchmarkRmi implements AutoCloseable {
      */
     static RemoteCallback lookup(int port) {
         try {
-            return (RemoteCallback)LocateRegistry.getRegistry(port).lookup(REMOTE_NAME);
+            return (RemoteCallback) LocateRegistry.getRegistry(port).lookup(REMOTE_NAME);
         } catch (IOException | NotBoundException e) {
             throw new RuntimeException(e);
         }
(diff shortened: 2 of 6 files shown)


ocadaruma:
example repository: https://github.com/ocadaruma/decaton-benchmark-comparison

kawamuray:
Before start reviewing this PR I'd like to ask few questions about the motivation.

> But currently, running custom Runner implementation cannot be done without modifying decaton repository itself since benchmark module is not published

What's the actual problem here?
I thought first publish `benchmark` module locally and develop a `Runner` with adding dependency to that, build a .jar containing `Runner` implementation and run `debm.sh` with classpath pointing to that jar file just works.
Since we don't expect many users to write a benchmark based on our `benchmark` module, this way just works for decaton devs.

> communicate ForkingExecution <-> benchmarker via RMI

I think this is a separate one from the above. and yeah.. I can understand that.
I think as a common protocol in unix systems, such debug information should go to stderr rather than stdout since stdout is expected to be used for I/O of essential data rather than random messages for users.
However in reality its usage differs among different commands and in fact I've had struggled to troubleshoot malfunctioning benchmark by stdout output issued by `profiler.sh` .. so I can agree with pursuing an alternative approach for communicating with child processes.
Do you think we have some possible alternatives to  RMI? how about creating a pipe manually?

ocadaruma:
At least we have to disable `noPublish` even for local publish I think? (refs https://docs.gradle.org/current/userguide/publishing_maven.html#publishing_maven:install)
> there's no need of modifying decaton repository

So, options here are:
- disable `noPublish` and publish to maven central as well as other modules
- or, introduce a mechanism to just publishing to maven-local (not to maven central)

I prefer former one because I think it's important to make benchmark reproduction in same condition easy as much as possible for everyone (not only Decaton dev), for reliability of benchmark-comparison (which is the main purpose of pluggable Runner).
If we see a benchmark code which depends on `benchmark` module published locally on benchmark executor's local machine, isn't it difficult to confirm it's valid and reproduce it in same condition?
> What's the actual problem here?

Then, why I separated `benchmark-api` rather than just publishing `benchmark` is just to avoid redistribution of many libs (we shade `benchmark` module entirely, so if we publish it doesn't it cause a lot of redistributions)

Yeah there are other options but I think RMI is appropriate this time as below reasons:
- I/F can be type-safe and easy-to understand (rather than bringing another I/F layer such like `Stage should be serialized as String for name() and each message should be delimited by newline`)
- OS-agnostic
- Code is simple enough I think?

Means named pipe? I'm not sure we do that in Java in simple while non OS dependent way (though our main target is *nix, isn't good to make it OS agnostic basically?)
> creating pipe manually

kawamuray:
> At least we have to disable noPublish even for local publish I think? 

Ah... I see missed that point.

>  I think it's important to make benchmark reproduction in same condition easy as much as possible for everyone (not only Decaton dev), for reliability of benchmark-comparison (which is the main purpose of pluggable Runner).
If we see a benchmark code which depends on benchmark module published locally on benchmark executor's local machine, isn't it difficult to confirm it's valid and reproduce it in same condition?

>  I think it's important to make benchmark reproduction in same condition easy as much as possible for everyone (not only Decaton dev),

Agree.

> If we see a benchmark code which depends on benchmark module published locally on benchmark executor's local machine, isn't it difficult to confirm it's valid and reproduce it in same condition?

but I'm not sure about this.
The compiled class file should be the same as long as the source code is the same, and I don't see how published artifact contributes to keep environment consistent among different people's machines.
I think people rather needs to be careful for aligning `javac` version, JVM version and machine setup to make reproducible results which is out of our control. (so reproducible benchmark is really not an easy problem in the first place)

Even though we publish `benchmark-api` artifact, people still needs to build `benchmark.jar` manually running build in this repository to get runtime because we don't want to ship shadow jar that includes all binaries of dependencies, which makes us need to list them all in regards to their license as you said. 

So while I don't see any major downsides of publishing artifact in maven central (maybe we need to clarify that we don't care about it's API compatibility or stability?), I don't see any benefits of doing it either.
Do you see any benefits besides the reproducibility?

> RMI

I see. Actually what I had in my mind is future stability of RMI since recently I never saw any applications using it for practical purpose TBH...
I saw https://openjdk.java.net/jeps/385 this kind of thing and bit concerned, but according to @ykubota there's no topic about deprecation of RMI itself so it should be fine.

> unix pipe

another idea would be to use object input/output stream if we have to listen to a port anyway, but well no much difference.

ocadaruma:
Hm that's right, then it sounds no much difference in terms of benchmark reproducibility.
> Even though we publish benchmark-api artifact, people still needs to build benchmark.jar manually

Then, benefit here will be just a gradle script matter I think?
If we don't publish it to maven central, we have to add a mechanism to allow only local publish somehow (which shouldn't be so hard but sounds bit tricky isn't it?)

I see. Thanks for confirming.
> but according to @ykubota there's no topic about deprecation of RMI itself so it should be fine

ocadaruma:
Ah, or just exclude by `publish.sh` (anyway it sounds bit awkward)
> we have to add a mechanism to allow only local publish somehow

kawamuray:
> we have to add a mechanism to allow only local publish somehow

It seems not so difficult with few more flags added? 
https://docs.gradle.org/current/userguide/publishing_customization.html#sec:publishing_maven:conditional_publishing

anyway if you're strong about that, I can agree with proceeding as-is. It doesn't make sense to me, but still isn't a big deal :)

ocadaruma:
Hm, I tested locally and confirmed controlling by `onlyIf` just simply worked.. (and running a test by depending on local-published shadowed `benchmark` also worked)

Then it seems fine with current repository structure. (maybe I'm too paranoid to minimize dependency which is exposed to custom runner project..)

So I'll change this PR goal to
- change IPC to RMI
- make `benchmark` project to published locally

Sounds fine?

kawamuray:
Ya, that sounds reasonable :)

ocadaruma:
Thanks, then ready to review

benchmark/src/main/java/com/linecorp/decaton/benchmark/ForkingExecution.java
```diff
@@ -50,67 +51,67 @@
...
+        BenchmarkResult result = execution.execute(config, stage -> {
+            try {
+                remoteCallback.onStage(stage);
+            } catch (RemoteException e) {
+                log.error("Failed to invoke remote callback", e);
+                throw new RuntimeException(e);
+            }
+        });
+
+        remoteCallback.onResult(mapper.writeValueAsString(result));
```
kawamuray, line 66: Any points to keep serializing result while we already have a different layer of object serialization? (RMI)

benchmark/src/main/java/com/linecorp/decaton/benchmark/ForkingExecution.java
```diff
@@ -50,67 +51,67 @@
...
+
             log.debug("Forking child process for run with: {}", cmd);
             process = new ProcessBuilder()
                     .command(cmd)
                     .redirectError(Redirect.INHERIT)
+                    .redirectOutput(Redirect.INHERIT)
                     .start();
+
+            Stage stage;
+            while ((stage = rmi.pollStage()) != Stage.FINISH) {
```
kawamuray, line 94: any points to put a queue in between for this? As we anyway forward the event to the another callback, would it be more natural to give the callback itself to Rmi from the beginning?

kawamuray (changes requested):

benchmark/README.md
```diff
@@ -13,11 +13,105 @@ Build a shadowJar:
...
+apply plugin: 'com.github.johnrengelman.shadow'
+```
+
+```sh
+cd /path/to/custom-runner
+
+./gradlew shadowJar
+```
+
+#### Build a shadowJar of the benchmark project:
```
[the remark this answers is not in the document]
ocadaruma: Oh that's true. I'll remove

benchmark/src/main/java/com/linecorp/decaton/benchmark/BenchmarkRmi.java
```diff
@@ -0,0 +1,166 @@
...
+ * Remote interface used for communication between forked execution and benchmark controller.
+ *
+ * {@link Stage} changes are notified through {@link BenchmarkRmi#pollStage()} rather than
+ * RMI callback directly which requires benchmark controller to be concurrent aware.
+ */
+@Slf4j
+class BenchmarkRmi implements AutoCloseable {
+    private static final String REMOTE_NAME = "BenchmarkRmi";
+
+    private final BlockingQueue<Stage> stageQueue = new LinkedBlockingQueue<>();
```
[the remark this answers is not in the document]
ocadaruma: That's because we have to specify capacity if we use ArrayBlockingQueue.

Yeah, in current implementation, there should be only 3 stage callback invocation at maximum (READY_WARMUP => READY => FINISH), but it sounds natural to use unbounded queue this time rather than specifying "sufficiently enough" capacity?

benchmark/src/main/java/com/linecorp/decaton/benchmark/ForkingExecution.java
ocadaruma: For this point: https://github.com/line/decaton/pull/68/files#diff-308fef37ee55388e364ad04c5992d6e2R40
wdyt?

benchmark/src/main/java/com/linecorp/decaton/benchmark/ForkingExecution.java
ocadaruma: Because we have to annotate `Serializable` for BenchmarkResult and all other inner classes.
Isn't better to serialize as JSON explicitly rather than forcing BenchmarkResult to have only Serializable fields (also for future extension)?

ocadaruma:
Checking if there's extra overhead with RMI:

- I ran RMI-version of benchmark jar several times with `--profile`, but I couldn't find RMI related stack. (stage callback will be invoked only 3 times per benchmark, so async-profiler didn't catch that? (because of sampling))

benchmark/src/main/java/com/linecorp/decaton/benchmark/BenchmarkRmi.java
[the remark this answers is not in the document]
kawamuray: I see. I prefer to put explicit capacity bound for all cases unless I really want it to potentially consume all available heap. so to me, Array with some sufficiently large cap (like 16 or 32) still makes sense, but it's fine to leave it as-is if it's backed by certain intention :)

benchmark/src/main/java/com/linecorp/decaton/benchmark/ForkingExecution.java
kawamuray: ugh... that's Java it is.
I see. as we need to keep it serializable into JSON anyway (for json format output), I think it's fine to keep this as-is.

benchmark/src/main/java/com/linecorp/decaton/benchmark/ForkingExecution.java
kawamuray: As for separating RMI thread with processing threads, it seems reasonable.

kawamuray (approved):
LGTM, nice!

PR #57, opened 2020-07-29
Integrate zipkin Brave into Decaton

yifu-yang:
Follow the issue #50，I try to intergrate the zipkin Brave into the Decaton.
If there is any issue or suggestion，please let me know.
Thanks!

Issue #50, opened 2020-06-26
Add distributed tracing support


ocadaruma:
- There are needs of distributed tracing support
- Currently Decaton doesn't provide a way to propagate tracing context to DecatonProcessor
  * refs. some implementation utilizes Kafka header: https://github.com/openzipkin/brave/tree/master/instrumentation/kafka-clients

ocadaruma:
fixed by https://github.com/line/decaton/pull/69

Issue #56, opened 2020-07-29
Fix flaky tests


ocadaruma:
- Some tests are occasionally failing
  * https://travis-ci.com/github/ocadaruma/decaton/jobs/358642260
  * https://travis-ci.com/github/line/decaton/jobs/365312195

PR #69, opened 2020-09-29
Allow client to support brave/zipkin distributed tracing in a non-intrusive way

m50d:
Based on #50 and the discussion in #57 . Not polished - wanted to discuss whether the architecture makes sense first.

Generalise the `LoggingContext` that's returned from `ProcessorContextImpl` to a `TracingContext`.

Allow it to be customised by a `TracingContextFactory`.

Define a brave-specific implementation that reads the zipkin trace information from the Kafka record headers in a separate `brave` module.


kawamuray:
Thanks for the PR!

I've looked the patch briefly and it looks good overall (as you correctly reflected my options at  #57  as you mentioned :) ).

One major point I'd like to discuss before start looking into the detail is the scope of tracing application.
Currently the tracing context is instantiated just before the `DecatonProcessor#process` starts, which should tell us the time taken between produce to right before the process, and time taken to complete processing.
However as I write in #57 I think when investigating a process latency issue we typically wants to know the exact time that the record is delivered to the consumer to distinguish latency among kafka message delivery (matter of kafka clients and broker), decaton internal (queuing and waiting for preceding tasks's complete) and processing time.
So I think we should support a bit more higher granularity for measurement. At last once at the time of consume, (Consumer returns the record) and right before the process.
For that, I think the tracing should be configured for `SubscriptionBuilder` and not for `ProcessorsBuilders`.

Also, some users for implements `DecatonProcessor` in async way (using `DeferredCompletion`) may want to measure the time they "complete" the task which seems to be not possible in current usage of brave.

Can you consider this point?

m50d:
> So I think we should support a bit more higher granularity for measurement. At last once at the time of consume, (Consumer returns the record) and right before the process.

Makes sense. In that case we also need to propagate the context from the `ProcessorSubscription`'s polling thread to the `ProcessorUnit`s executor. I'm not sure how decoupled this will actually be, but I'll try to make an interface that makes sense.
 
> Also, some users for implements `DecatonProcessor` in async way (using `DeferredCompletion`) may want to measure the time they "complete" the task which seems to be not possible in current usage of brave.

The processor implementation can start its own spans, which will have the current span as a parent (since it's in context on the current thread). So I don't think we need to do anything at the decaton level here (and indeed I think we don't need to start a span around the process call either - if the processor wants this to be a separate span it can start one itself). That said I do think it makes more sense to close the span when the processresult is completed rather than when the call to push returns though.

Will rework.

kawamuray:
> it makes more sense to close the span when the processresult is completed rather than when the call to push returns though.

Indeed. Decaton already has two concept for tracking task's processing duration - process time and complete time, as they're exposed independently in metrics. Process time corresponds to the time that the `process` method took till it returns and complete time corresponds to the time the task's offset is marked as completed, so complete time includes process time. I think spans can follow the same structure for this. The span measuring task's complete time (starts right before `process` call, finish when the process result's marked as completed) and it has one or more children where's the first one is "process time" - the duration of `process` method call.
As you said users can add one or more child span under the "complete time" span or under the "process time" span depending on whether its executed synchronously in process method or not, so they're likely get the view on zipkin UI with no gap between spans if they design their own spans correctly.

m50d:
>  I think spans can follow the same structure for this. The span measuring task's complete time (starts right before process call, finish when the process result's marked as completed) and it has one or more children where's the first one is "process time" - the duration of process method call.

> As you said users can add one or more child span under the "complete time" span or under the "process time" span depending on whether its executed synchronously in process method or not, so they're likely get the view on zipkin UI with no gap between spans if they design their own spans correctly.

I can see that many users will want that kind of structure, but I think for some users two separate spans might be more complex than necessary. So I don't think we actually need to provide the "process time" span at decaton framework level - if a user wants a span that covers the execution of their `process` method, they can define a `process` method that starts a (child) span at the start and closes it just before returning (or some more complex structure of child spans, some or all of which may be async). But if the user prefers to see a single span per record for simplicity then I'd like to support that use case as well.

processor/src/main/java/com/linecorp/decaton/processor/runtime/SubscriptionBuilder.java
```diff
@@ -126,6 +131,16 @@ public SubscriptionBuilder enableRetry(RetryConfig config) {
         return this;
     }
 
+    /**
+     * @param tracingProvider Add tracing provider that will be called to trace the execution of processing
+     * each record.
+     * @return updated instance of {@link SubscriptionBuilder}.
+     */
+    public SubscriptionBuilder enableTracing(TracingProvider tracingProvider) {
+        this.tracingProvider = tracingProvider;
```
kawamuray, line 140: please filter this with `Objects.requireNonNull` since null argument for `enableXX` just doesn't make sense.

kawamuray (changes requested):

kawamuray:
> if a user wants a span that covers the execution of their process method, they can define a process method that starts a (child) span at the start and closes it just before returning (or some more complex structure of child spans, some or all of which may be async)

Yeah, that's true. At the same time one of the key benefits that a library provides is a good default that helps many users from needing to write same boilerplates.

As you said, especially for sync `process` implementations having two parts - process time and complete time  - is almost pointless since they both shows exactly the same durations (I guess two same bars up and down on zipkin UI?)
At the same time, we saw many users use deferred completion (async `process`) with relatively lightweight synchronous work in process method, which in case they want to add another span for measuring synchronous part in process I guess.

So a good default depends on how harmful it is to have both process and complete measurements for sync type process users.

> I can see that many users will want that kind of structure, but I think for some users two separate spans might be more complex than necessary. 

Can you explain bit more about how is it bad for users to have 2 spans? Is it about two, overlapping bars on UI?

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -353,9 +359,17 @@ private void pollOnce(Consumer<String, byte[]> consumer) {
 
         timer = Utils.timer();
         records.forEach(record -> {
+            final Closeable trace = null != tracingProvider ? tracingProvider.startTraceFor(record) : null;
```
ocadaruma, line 362: Isn't better to prepare something like `NoopTracingProvider` rather than null checking `tracingProvider` in many places?

brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTracingProvider.java
```diff
@@ -0,0 +1,49 @@
...
+public class BraveTracingProvider implements TracingProvider {
+    private final KafkaTracing kafkaTracing;
+
+    public BraveTracingProvider(KafkaTracing kafkaTracing) {
+        this.kafkaTracing = kafkaTracing;
+    }
+
+    @Override
+    public Closeable startTraceFor(ConsumerRecord<?, ?> record) {
+        final Span span = kafkaTracing.nextSpan(record).name("process").start();
```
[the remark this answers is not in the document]
m50d: The convention seems to be short names (`post` and `redis` are other examples), but `decaton.process` sounds reasonable. Will do that.

processor/src/it/java/com/linecorp/decaton/processor/RetryQueueingTest.java
```diff
@@ -89,6 +92,8 @@ public void testRetryQueuing() throws Exception {
                 .retryConfig(RetryConfig.builder()
                                         .retryTopic(retryTopic)
                                         .backoff(Duration.ofMillis(10))
+                                        .producerSupplier(config -> new TestTracingProducer(
```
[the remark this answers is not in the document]
m50d: The retry producer needs to propagate the trace id, since the test asserts that it's set correctly at the point where the message is processed (even on retry). In the real case a zipkin tracing kafka producer will do this (I put a note about this in the comment on `TracingProvider`).

testing/src/main/java/com/linecorp/decaton/testing/KafkaAdmin.java
```diff
@@ -63,10 +63,12 @@ public void createTopic(String topicName, int numPartitions, int replicationFact
     }
 
     public void deleteTopics(String... topicNames) {
-        try {
-            adminClient.deleteTopics(Arrays.asList(topicNames)).all().get();
-        } catch (Exception e) {
-            throw new RuntimeException(e);
+        if (!EmbeddedKafkaCluster.IS_WINDOWS) {
```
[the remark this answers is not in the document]
m50d: > https://github.com/line/decaton/pull/69/files#diff-2d8a3a3355ba2b4c1d8f5e4395d91804R71 isn't this change enough to prevent kafka to perform physical deletion, so we can bypass the problem described in jira?

That change means the embedded kafka cluster throws an exception instead of performing deletion. So we still need to avoid calling the delete method somehow.

> Even if the target cluster is a locally running embedded cluster, I think to avoid failing build due to related tests on windows its better to skip the related test itself (conditionally ignore) on windows rather than disabling only the topic delete because the testing logic might depends on topic deletion. It's not desirable but at least we can still catch an issue at CI if the change introduced any undesired side effect.

In principle I agree with you, but the only place this method is actually called is as a "cleanup"-style deletion. So it seems a shame to disable all our tests on windows when actually they could run fine. How about if I add a flag e.g. `failSilently`, that can be `false` for cleanup and `true` if the test needs the deletion to happen as part of its semantics (in which case the test would have to be skipped on Windows).

testing/src/main/java/com/linecorp/decaton/testing/processor/ProcessingGuarantee.java
```diff
@@ -74,7 +75,11 @@
         /**
          * Tasks which have same key never be processed simultaneously
          */
-        SERIAL_PROCESSING;
+        SERIAL_PROCESSING,
+        /**
+         * Trace IDs will be propagated to task processing threads if a suitable {@link TracingProvider} is set
+         */
+        TRACE_PROPAGATION;
```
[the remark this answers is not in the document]
m50d: Makes sense. Will do.

m50d:
> Can you explain bit more about how is it bad for users to have 2 spans? Is it about two, overlapping bars on UI?

Yes, and also increasing the storage volume (which in our case has been a significant concern). Maybe an annotation within the span would be the right way to represent this in Zipkin.

> At the same time, we saw many users use deferred completion (async process) with relatively lightweight synchronous work in process method, which in case they want to add another span for measuring synchronous part in process I guess.

At least in the zipkin case, such a user already has to add some boilerplate to propagate the context onto that async task, so I don't think it's any more overhead for them to start a new span at the same time.

I had wondered about implementing a parent class (`TracedDecatonProcessor` or some such) that starts the spans in the right places. But I think that would only be worthwhile for tracing implementations that automatically propagate the context across async calls, of which the only one I know about is Kamon.

processor/src/main/java/com/linecorp/decaton/processor/runtime/TracingProvider.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Interface for distributed tracing implementations that track traces via Kafka {@link ConsumerRecord}s
+ * (typically through Kafka headers) e.g. Zipkin, OpenTracing, Kamon.
+ * Note that this will usually need to be paired with a suitable {@link KafkaProducerSupplier} on the producer
+ * side to write those Kafka headers.
+ * In particular, if you want retries to be shown as part of the same (distributed) trace, make sure to supply
+ * such a {@link KafkaProducerSupplier} in your {@link RetryConfig}.
+ */
+public interface TracingProvider {
```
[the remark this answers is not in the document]
m50d: I went through the same thought process myself. If we want to pass a customisable `TracingContext` down into the processing implementation, then it would have to be carried somewhere (e.g. `ProcessingContext`) and would then have to be exposed as a type parameter on that, and on the `DecatonProcessor` as well - it seems too intrusive for something that's just meant to be tracing. And we would still need to do some work to make the trace context "current" on the thread that's executing the processor, so we'd either need to add more to this interface (or at least provide some non-obvious guarantees about which thread will call `onStage` with which value) or the user would have to write boilerplate in their processor.

I believe all the major tracing implementations use a thread-local context - at least Zipkin, OpenTracing, New Relic and Kamon do. So I think it's better to acknowledge that and design an interface that will work well for the tracing libraries that we're likely to support. Unless there's actually a non-thread-local-based tracing library that we want to implement support for?

processor/src/it/java/com/linecorp/decaton/processor/RetryQueueingTest.java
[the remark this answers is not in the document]
kawamuray: I see. Please check that in to-be added tracing integration test rather than modifying existing RetryQueueingTest.

kawamuray (changes requested):

testing/src/main/java/com/linecorp/decaton/testing/KafkaAdmin.java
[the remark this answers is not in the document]
kawamuray: Okay that sounds like an option 👍

processor/src/main/java/com/linecorp/decaton/processor/runtime/TracingProvider.java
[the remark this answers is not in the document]
kawamuray: > So I think it's better to acknowledge that and design an interface that will work well for the tracing libraries that we're likely to support. 

Yeah, that's really a good point.
However, my concern is not only about applicability of this interface for any possible instrumentation libraries, but also maintainability of the protocol between decaton core and drivers.
By making this interface to implicitly relying on Thread context, that would makes decaton core impl need to take care of it when considering it's internal threading model. As you did in propagateXX method, for every place that decaton switches the thread it needs to ensure calling a method to propagate context to the next thread which sounds a fragile implication that can be easily broken when decaton's internal architecture changes (even if tests are supposed to catch it).
I'm also not sure if propagation always happen through Runnable interface. e.g, previous decaton implementation used a queue with a task type which will be received by a worker (a already running Runnable) on the other side.
So I think it's the best if we can make the list of information that are exchanged between decaton core and tracing driver always explicit on the interface, rather than expecting the thread local to hold particular information at a place.

> would then have to be exposed as a type parameter on that, and on the DecatonProcessor as well - it seems too intrusive for something that's just meant to be tracing. 

and this is true. So I think a good compromise is to let decaton processor code (user code) to extract their expected tracing type out of interface type by casting, if necessary.
While I don't prefer to introduce such dynamic casting in defined interface as possible as we can, I think it is essentially not much different from extracting context from the thread local. Both relies on a particular information that is not certain if it really exists (expected type vs potentially missing thread local field) from the processor's limited point of view. I mean - that's what they cannot be benefited by static checking.

``java
tc = (BraveTracingContext) context.tracingContext(); // or context.<BraveTracingContext>tracingContextAs() ?
vs
Tracing.getCurrentContext()
```

Btw as long as the instrumentation library uses thread local for holding a context, I think users won't need to access specific TracingProvider to instrument their synchronous processing parts? While I'm against to use thread local for bringing around the context inside the decaton core, I think it's good to setup (inside the driver, at `onStage`) thread local around calling `process` so users can just use the library in regular way. This should be fine because it's a protocol between each specific drivers and user that selects to use it and decaton core doesn't care about it (it just calls onStage).

> still need to do some work to make the trace context "current" on the thread that's executing the processor, so we'd either need to add more to this interface (or at least provide some non-obvious guarantees about which thread will call onStage with which value) 

This sounds fine though? "onStage(X) is called at place where X starts, by a thread that actually executes X" sounds like a reasonable contract to follow. We just don't want to let drivers to assume that the thread executes X and thread executes Y are contagious and can always "propagate" tracing context through thread local.

**Comment by kawamuray** [comment]
> At least in the zipkin case, such a user already has to add some boilerplate to propagate the context onto that async task, so I don't think it's any more overhead for them to start a new span at the same time.

Okay. Let's go with single span for process (complete) part for now. We can easily add one another if we got demands for it :)

File: processor/src/main/java/com/linecorp/decaton/processor/runtime/TracingProvider.java (hunk 7 above)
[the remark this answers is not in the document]
Comment (m50d) [reply processor/src/main/java/com/linecorp/decaton/processor/runtime/TracingProvider.java]: Ah, I see. Yeah, that makes sense. I've reworked the interface with a similar design to what you suggest; please take a look and see what you think.

File: brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTracingProvider.java [hunk 8]
```diff
@@ -0,0 +1,57 @@
...
+    private static final ThreadLocal<Scope> CURRENT_SCOPE = new ThreadLocal<>();
+    private final KafkaTracing kafkaTracing;
+
+    public BraveTracingProvider(KafkaTracing kafkaTracing) {
+        this.kafkaTracing = kafkaTracing;
+    }
+
+    @Override
+    public Span traceFor(ConsumerRecord<?, ?> record) {
+        return kafkaTracing.nextSpan(record).name("decaton.process").start();
```
kawamuray, line 36: "decaton.process" sounds like to correspond for `DecatonProcessor#process` method call rather than the entire flow in decaton consumer doesn't it?
To me below naming are the most natural:
* traceFor => decaton.consume (or just "decaton"?)
* processingStart => decaton.process(or).start
* processingReturn => decaton.process(or).return

WDYT?

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessingContextImpl.java
```diff
@@ -77,6 +77,11 @@ public String subscriptionId() {
         return subscriptionId;
     }
 
+    @Override
+    public Object trace() {
```
kawamuray, line 81: WRT https://github.com/line/decaton/pull/69/files#diff-0643ff76b830bb4b4a3eeb61dc5e1563R54 this comment, the return type can be at least `TracingContext` so that users will know that the object is at least an implementation of `TracingContext` ?

Or if we assume all existing instrumentation libraries to rely on thread local, and all users need in DecatonProcessor is that, then we might not need to expose anything through `ProcessingContext` so far.

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -346,6 +353,11 @@ public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
         }
     }
 
+    private <T> TraceAndCompletion<T> startTrace(TracingProvider<T> provider, ConsumerRecord<?, ?> record) {
```
kawamuray, line 356: can be a `static`

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessPipeline.java
```diff
@@ -109,6 +109,7 @@ public void scheduleThenProcess(TaskRequest request) throws InterruptedException
         Timer timer = Utils.timer();
         CompletableFuture<Void> processResult;
         try (LoggingContext ignored = context.loggingContext()) {
+            scope.tracingProvider().unsafeProcessingStart(request.trace());
```
kawamuray, line 112: This and below `xxReturn` call are better to be surrounded by try-catch with logging?
The implementation is blackbox from decaton core and while we expect them to never throw, if it happens, I feel like we shouldn't let decaton to abort just because of tracing function flaw.

brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTracingProvider.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Override
+    public void processingStart(Span trace) {
+        CURRENT_SCOPE.set(
+                kafkaTracing.messagingTracing().tracing().currentTraceContext().newScope(trace.context()));
+        trace.annotate("decaton.pipeline.start");
```
kawamuray, line 43: I'm not enough experienced with zipkin/brave so not attempting to doubt something but just to learn: what's the criteria of deciding whether to use span and annotation?

kawamuray (changes requested):
Thanks for redesigning interface, look better :) just few more points.

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -346,6 +353,11 @@ public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
...
+        T trace = provider.traceFor(record);
+        return new TraceAndCompletion<>(trace, () -> provider.processingCompletion(trace));
```
ocadaruma, line 358: For this point: this call should be surrounded by try-catch too I think https://github.com/line/decaton/pull/69/files#diff-d8bf175344f574cb7bc4d55debac19f5R80

ocadaruma (commented):
added minor comment, but almost lgtm except kawamuray's points.

docs/tracing.adoc
```diff
@@ -0,0 +1,26 @@
+= Tracing
+:base_version: 0.0.42
+:modules: brave,processor
+
+Decaton can integrate with distributed tracing frameworks so that you can associate the processing of a message
+with the component that originally sent that message. This is implemented via the
+ link:../processor/src/main/java/com/linecorp/decaton/processor/runtime/TracingProvider.java[TracingProvider]
```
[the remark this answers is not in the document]
m50d: No idea. Will fix.

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessPipeline.java
m50d: Point taken. Will do.

processor/src/main/java/com/linecorp/decaton/processor/runtime/TracingProvider.java
```diff
@@ -0,0 +1,86 @@
...
+     * Implementations may wish to associate the passed trace with the current thread
+     * (e.g. using a {@link ThreadLocal}) for use during {@link DecatonProcessor} execution
+     * @param trace returned from traceFor for the original Kafka record
+     */
+    void processingStart(T trace);
+
+    /**
+     * Entry point to processingStart. It is not recommended to override this method.
+     */
+    default void unsafeProcessingStart(Object trace) {
```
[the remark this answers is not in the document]
m50d: I was thinking of keeping close and open close together, and roughly corresponding to `Tracing`/`Span` in Brave, but yeah that sounds better. Let me rework it.

brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTracingProvider.java
m50d: Will rename the span to `decaton`.
I didn't want to say `decaton.process.start` because I realised this will actually only call `start`/`return` once for the whole `ProcessPipeline`, which might actually include multiple processors.

brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTracingProvider.java
m50d: I'm not an expert either, but I think the big difference is that a span has its own tags, whereas an annotation is basically just a timestamp on an existing span. So e.g. if we wanted to set some kind of processor id as a tag then we'd need to make a span for each processor.

kawamuray: Thanks for explanation :)

> if we wanted to set some kind of processor id as a tag then we'd need to make a span for each processor.

and this made me to came up with another question; shouldn't we pass 'subscriptionId' to `traceFor` method so that the driver can set it as a tag in span? otherwise how can one distinguish which decaton instance does the trace come from?

testing/src/main/java/com/linecorp/decaton/testing/KafkaAdmin.java
```diff
@@ -62,11 +62,13 @@ public void createTopic(String topicName, int numPartitions, int replicationFact
...
     }
 
-    public void deleteTopics(String... topicNames) {
+    public void deleteTopics(boolean ignoreFailures, String... topicNames) {
         try {
             adminClient.deleteTopics(Arrays.asList(topicNames)).all().get();
         } catch (Exception e) {
-            throw new RuntimeException(e);
+            if (!ignoreFailures) {
+                throw new RuntimeException(e);
```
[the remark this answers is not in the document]
ocadaruma: deleteTopics doesn't but `.get()` does. so this conversion is necessary

processor/src/main/java/com/linecorp/decaton/processor/ProcessingContext.java
```diff
@@ -48,6 +49,12 @@
      */
     LoggingContext loggingContext();
 
+    /**
+     * @return The trace (returned by {@link TracingProvider}) associated with the task currently being
+     * processed. May be null.
+     */
+    Object trace();
```
ocadaruma, line 56: Can't be `TraceHandle` ?

brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTracingProvider.java
```diff
@@ -0,0 +1,57 @@
...
+    private final KafkaTracing kafkaTracing;
+
+    public BraveTracingProvider(KafkaTracing kafkaTracing) {
+        this.kafkaTracing = kafkaTracing;
+    }
+
+    @Override
+    public TraceHandle traceFor(ConsumerRecord<?, ?> record) {
+        final Span span = kafkaTracing.nextSpan(record).name("decaton").start();
+        return new TraceHandle() {
```
ocadaruma, line 37: For this point: https://github.com/line/decaton/pull/69/files#r500168178

Developers might cast `ProcessingContext#trace()` to specific TraceHandle to use impl-specific information (like brave's `Span` (though currently Span isn't exposed and I think it's not necessary to expose for now)), so shouldn't we define something like `BraveTraceHandle` rather than instantiating anonymous class ?

processor/src/main/java/com/linecorp/decaton/processor/ProcessingContext.java
m50d: It's no longer actually needed - I'm thinking about removing it unless we have a use case for it? WDYT?

brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTracingProvider.java
m50d: If we don't expose the `Span` then I don't think it makes any sense to create a specific class either, since there's nothing a user can actually do with that `BraveTraceHandle`? My instinct is to keep everything hidden unless there's a use case for exposing the impl-specific implementation, but I'm happy to expose it if you think that's better.

brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTracingProvider.java
m50d: Good point - will add that.

testing/src/main/java/com/linecorp/decaton/testing/KafkaAdmin.java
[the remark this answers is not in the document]
kawamuray: oh I missed that, thanks.

processor/src/main/java/com/linecorp/decaton/processor/ProcessingContext.java
kawamuray: +1. we can add it when it become necessary w/o breaking compatibility.

brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTracingProvider.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    public TraceHandle traceFor(ConsumerRecord<?, ?> record, String subscriptionId) {
+        final Span span = kafkaTracing.nextSpan(record).name("decaton").tag("subscriptionId", subscriptionId)
+                                      .start();
+        return new TraceHandle() {
+            @Override
+            public void processingStart() {
+                CURRENT_SCOPE.set(
```
kawamuray, line 41: It seems that using thread local is not meaningful anymore? Just setting a scoped variable (or field) seems to be enough and better.

docs/retry-queueing.adoc
```diff
@@ -1,5 +1,5 @@
 = Retry Queuing
-:base_version: 0.0.40
+:base_version: 0.0.42
```
kawamuray, line 2: please also bump other docs that has `:modules:` decl for `processor` to make a build passes.

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -44,8 +44,10 @@
 import com.linecorp.decaton.processor.SubscriptionStateListener;
 import com.linecorp.decaton.processor.metrics.Metrics;
 import com.linecorp.decaton.processor.metrics.Metrics.SubscriptionMetrics;
+import com.linecorp.decaton.processor.runtime.TracingProvider.TraceHandle;
 import com.linecorp.decaton.processor.runtime.Utils.Timer;
 
+import lombok.Value;
```
kawamuray, line 50: unused

processor/src/main/java/com/linecorp/decaton/processor/runtime/TaskRequest.java
```diff
@@ -34,17 +35,20 @@
...
+    private final TraceHandle trace;
+    @ToString.Exclude
     private byte[] rawRequestBytes;
 
     TaskRequest(TopicPartition topicPartition,
                 long recordOffset,
                 DeferredCompletion completion,
                 String key,
-                byte[] rawRequestBytes) {
+                TraceHandle trace, byte[] rawRequestBytes) {
```
kawamuray, line 46: please wrap line when arguments are taking one-per-line each style

processor/src/main/java/com/linecorp/decaton/processor/runtime/TracingProvider.java
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Interface for distributed tracing implementations that track traces via Kafka {@link ConsumerRecord}s
+ * (typically through Kafka headers) e.g. Zipkin, OpenTracing, Kamon.
+ * Note that this will usually need to be paired with a suitable {@link KafkaProducerSupplier} on the producer
+ * side to write those Kafka headers.
+ * In particular, if you want retries to be shown as part of the same (distributed) trace, make sure to supply
+ * such a {@link KafkaProducerSupplier} in your {@link RetryConfig}.
+ */
+@FunctionalInterface
```
kawamuray, line 34: AFAIU, putting` @FunctionInterface` indicates that we intend to keep this interface to have just one required function to implement, which is not a case for this I think.

kawamuray (changes requested):

benchmark/src/main/java/com/linecorp/decaton/benchmark/TemporaryTopic.java
```diff
@@ -41,7 +41,7 @@ public static TemporaryTopic create(String bootstrapServers, String topic) {
     @Override
     public void close() throws Exception {
         log.info("Cleaning up topic {}", topic);
-        admin.deleteTopics(topic);
+        admin.deleteTopics(true, topic);
```
kawamuray, line 44: I think this should be false because benchmark could run against a custom cluster (that could be an actually servicing cluster) depending on configuration rather than a local embedded cluster and in case we must clean it up.

kawamuray:
Except those minor points now this looks good overall to me.
Please check comments once again since there are some comments folded by github and still aren't addressed (like unused import)

processor/src/main/java/com/linecorp/decaton/processor/runtime/TracingProvider.java
m50d: I'd expect this interface will always consist of just `traceFor` and everything else will be on the `TraceHandle`.

kawamuray: yeah, might be. but people can still use lambda for this class regardless to whether its annotated by `@FunctionalInterface` or not, making me to feel there's no benefit to put it except for the limited cases that we strongly intend it to be an argument for functional-style interfaces.
it's just a way of my thinking so feel free to leave it if you like. (we'd just remove it when we need to add an another method in this interface)

processor/src/main/java/com/linecorp/decaton/processor/ProcessingContext.java
ocadaruma: I see. understood. then let's just remove for now

brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTracingProvider.java
ocadaruma: I see.
Regarding this: https://github.com/line/decaton/pull/69#discussion_r502987749 , we don't provide a way to access underlying trace for now so anonymous instantiation should be fine.

processor/src/main/java/com/linecorp/decaton/processor/runtime/TracingProvider.java
m50d: Ah, I hadn't realised that. To be honest I was mainly going by IntelliJ's warning. Will remove.

m50d:
Hi,

I don't understand the Travis build failure. Since I couldn't reproduce it locally I've rebased the branch to the minimum change that causes the failure. It looks like just adding the trace ID header in `ProcessorTestSuite` causes these two test failures. Are either of you able to reproduce them? Or is there any way to see what's happening on Travis in more detail?

Many thanks,
Mickey

kawamuray:
should be #56 ...

it shouldn't keep failing forever though. 
6733994 got it passed apparently.
Once you finalize your PR, please just check the test results. If it's difficult to make it pass with 1 or 2 retries (due to those flaky ones) we'll just proceed to merge it disregarding them for now.

m50d:
> should be #56 ...
> 
> it shouldn't keep failing forever though.
> [6733994](https://github.com/line/decaton/commit/67339943d85fb9be5b3331fdd2ad435e95a39802) got it passed apparently.
> Once you finalize your PR, please just check the test results. If it's difficult to make it pass with 1 or 2 retries (due to those flaky ones) we'll just proceed to merge it disregarding them for now.

Ah, I see. Thanks.
I'm happy with the PR (and I think I've addressed all your and @ocadaruma 's comments?).
It looks like I can't re-run the Travis test myself since I don't have write access to this repository?

ocadaruma (approved):
LGTM

kawamuray (approved):
LGTM 👍 

Thanks for contributing a great feature!


All CI failures are caused by #56 and are not related to this patch.

PR #70, opened 2020-10-13
Fix flaky tests

### LINKED ISSUE #56: Fix flaky tests

kawamuray:
Attempt to fix #56 

* `testRetryQueuing` => not very sure but suspecting CI server's resource simply lacks to complete the workload
* `testMetricsCleanup` => likely involved by the above test's failure. timed out test likely gets terminated w/o completing finally block and it'll laves garbage in metrics registry on shared JVM.
* `testRaceConditionOnFlush` => no idea yet. occurrence is too rare and not a subject of this PR.

### DIFF
--- a/processor/src/it/java/com/linecorp/decaton/processor/RetryQueueingTest.java
+++ b/processor/src/it/java/com/linecorp/decaton/processor/RetryQueueingTest.java
@@ -81,6 +81,7 @@ public class RetryQueueingTest {
         //   * after retried (i.e. retryCount() > 0), no more retry
         ProcessorTestSuite
                 .builder(rule)
+                .numTasks(1000)
                 .configureProcessorsBuilder(builder -> builder.thenProcess((ctx, task) -> {
                     if (ctx.metadata().retryCount() == 0) {
                         ctx.retry();
--- a/processor/src/it/java/com/linecorp/decaton/processor/metrics/MetricsTest.java
+++ b/processor/src/it/java/com/linecorp/decaton/processor/metrics/MetricsTest.java
@@ -14,7 +14,7 @@
  * under the License.
  */
 
-package com.linecorp.decaton.processor;
+package com.linecorp.decaton.processor.metrics;
 
 import static java.util.Collections.emptyList;
 import static org.junit.Assert.assertEquals;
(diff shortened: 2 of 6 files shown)


ocadaruma (commented):
Thanks for fixing. Changes look fine.

About `testRaceConditionOnFlush` failure, I could reproduce that by running following test case several times.

Do you have any ideas how this can happen?

```java
    @Test(timeout = 60000)
    public void testTasksCount() throws Exception {
        for (int i = 0; i < 10000; i++) {
            CountDownLatch schedulePassed = new CountDownLatch(1);
            CountDownLatch firstFlushComplete = new CountDownLatch(1);
            ScheduledThreadPoolExecutor scheduledExecutor = new ScheduledThreadPoolExecutor(1);

            AtomicInteger flushCount = new AtomicInteger();

            Runnable flushTask = () -> {
                int count = flushCount.getAndIncrement();
                if (count == 0) {
                    try {
                        schedulePassed.await();
                    } catch (InterruptedException e) {
                        throw new RuntimeException(e);
                    }
                    firstFlushComplete.countDown();
                }
            };
            Runnable process = () -> {
                scheduledExecutor.schedule(flushTask, 1L, TimeUnit.MILLISECONDS);
                schedulePassed.countDown();
                try {
                    firstFlushComplete.await();
                } catch (InterruptedException e) {
                    throw new RuntimeException(e);
                }
            };

            process.run();
            process.run();
            assertEquals(2, scheduledExecutor.getTaskCount());

            scheduledExecutor.shutdown();
            scheduledExecutor.awaitTermination(Long.MAX_VALUE, TimeUnit.MILLISECONDS);
        }
    }
```

kawamuray:
> Do you have any ideas how this can happen?

The only thing I could imagine is a race in `ScheduledThreadPoolExecutor#getTaskCount()`. as its javadoc says it returns approximate count of scheduled tasks and that could be not exact.
Reading its implementation briefly, it seems to hold the counter in 2 dimensions, one as a global counter in scheduler and one for each worker threads.
It returns the sum of two counters, potentially incrementing the count when the worker thread is locked, which happens not only when the worker thread is doing job but also when an idle thread gets terminated.
It's still bit weird as I see those places properly holding a lock global in scheduler but I'd like to try fixing it by terminating the thread before getting the count at first so we'd know if this is the case.

ocadaruma (approved):
LGTM. Thanks!

PR #71, opened 2020-10-15
Make "child" traces around each processor within the ProcessingPipeline

m50d:
To help understand performance in more detail, create a "child" `TraceHandle` around the individual processor.

### DIFF
--- a/brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveProcessorTraceHandle.java
+++ b/brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveProcessorTraceHandle.java
@@ -16,22 +16,28 @@
 
 package com.linecorp.decaton.processor.runtime;
 
-import com.linecorp.decaton.processor.TracingProvider.TraceHandle;
+import com.linecorp.decaton.processor.TracingProvider.ProcessorTraceHandle;
 
 import brave.Span;
 import brave.messaging.MessagingTracing;
+import brave.propagation.CurrentTraceContext.Scope;
 
-class BraveTraceHandle implements TraceHandle {
-    protected final MessagingTracing messagingTracing;
-    protected final Span span;
+final class BraveProcessorTraceHandle extends BraveTraceHandle implements ProcessorTraceHandle {
+    private Scope scope;
 
-    BraveTraceHandle(MessagingTracing messagingTracing, Span span) {
-        this.messagingTracing = messagingTracing;
-        this.span = span;
+    BraveProcessorTraceHandle(MessagingTracing messagingTracing, Span span) {
+        super(messagingTracing, span);
     }
 
     @Override
-    public void processingCompletion() {
-        span.finish();
+    public void processingStart() {
+        scope = messagingTracing.tracing().currentTraceContext().newScope(span.context());
+    }
+
+    @Override
+    public void processingReturn() {
+        span.annotate("return");
+        scope.close();
     }
 }
+
--- a/brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTraceHandle.java
+++ b/brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTraceHandle.java
@@ -17,7 +17,7 @@
 package com.linecorp.decaton.processor.runtime;
 
 import com.linecorp.decaton.processor.DecatonProcessor;
-import com.linecorp.decaton.processor.runtime.TracingProvider.TraceHandle;
+import com.linecorp.decaton.processor.TracingProvider.TraceHandle;
 
 import brave.Span;
 import brave.messaging.MessagingTracing;
--- a/brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTraceHandle.java
+++ b/brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTraceHandle.java
@@ -36,12 +36,11 @@ class BraveTraceHandle implements TraceHandle {
     @Override
     public void processingStart() {
         scope = messagingTracing.tracing().currentTraceContext().newScope(span.context());
-        span.annotate("decaton.start");
     }
 
     @Override
     public void processingReturn() {
-        span.annotate("decaton.return");
+        span.annotate("return");
         scope.close();
     }
 
--- a/brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTraceHandle.java
+++ b/brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTraceHandle.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2020 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.decaton.processor.runtime;
+
+import com.linecorp.decaton.processor.DecatonProcessor;
+import com.linecorp.decaton.processor.runtime.TracingProvider.TraceHandle;
+
+import brave.Span;
+import brave.messaging.MessagingTracing;
+import brave.propagation.CurrentTraceContext.Scope;
+
+class BraveTraceHandle implements TraceHandle {
+    private final MessagingTracing messagingTracing;
+    private final Span span;
+    private Scope scope;
+
+    public BraveTraceHandle(MessagingTracing messagingTracing, Span span) {
+        this.messagingTracing = messagingTracing;
+        this.span = span;
+    }
+
+    @Override
+    public void processingStart() {
+        scope = messagingTracing.tracing().currentTraceContext().newScope(span.context());
+        span.annotate("decaton.start");
+    }
+
+    @Override
+    public void processingReturn() {
+        span.annotate("decaton.return");
+        scope.close();
+    }
+
+    @Override
+    public void processingCompletion() {
+        span.finish();
+    }
+
+    @Override
+    public TraceHandle childFor(DecatonProcessor<?> processor) {
+        final Span childSpan = messagingTracing.tracing().tracer().newChild(span.context())
+                                               .name(processor.getClass().getSimpleName())
+                                               .start();
+        return new BraveTraceHandle(messagingTracing, childSpan);
+    }
+}
--- a/brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTraceHandle.java
+++ b/brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTraceHandle.java
@@ -28,7 +28,7 @@ class BraveTraceHandle implements TraceHandle {
     private final Span span;
     private Scope scope;
 
-    public BraveTraceHandle(MessagingTracing messagingTracing, Span span) {
+    BraveTraceHandle(MessagingTracing messagingTracing, Span span) {
         this.messagingTracing = messagingTracing;
         this.span = span;
     }
@@ -53,7 +53,7 @@ class BraveTraceHandle implements TraceHandle {
     @Override
     public TraceHandle childFor(DecatonProcessor<?> processor) {
         final Span childSpan = messagingTracing.tracing().tracer().newChild(span.context())
-                                               .name(processor.getClass().getSimpleName())
+                                               .name(processor.name())
                                                .start();
         return new BraveTraceHandle(messagingTracing, childSpan);
     }
--- a/docs/consuming-any-data.adoc
+++ b/docs/consuming-any-data.adoc
@@ -1,6 +1,6 @@
 Consuming Arbitrary Topic
 =========================
-:base_version: 0.0.43
+:base_version: 0.0.43-SNAPSHOT
 :modules: common,protocol,processor
 
 This document guides you how to consume and process topics containing records not consists of Decaton's protocol (not produced by DecatonClient) using Decaton processors.
--- a/docs/consuming-any-data.adoc
+++ b/docs/consuming-any-data.adoc
@@ -1,6 +1,6 @@
 Consuming Arbitrary Topic
 =========================
-:base_version: 0.0.42
+:base_version: 0.0.43
 :modules: common,protocol,processor
 
 This document guides you how to consume and process topics containing records not consists of Decaton's protocol (not produced by DecatonClient) using Decaton processors.
--- a/processor/build.gradle
+++ b/processor/build.gradle
@@ -8,6 +8,7 @@ buildscript {
 }
 
 apply plugin: "me.champeau.gradle.jmh"
+apply plugin: "java-test-fixtures"
 
 dependencies {
     compile project(":common")
@@ -16,6 +17,9 @@ dependencies {
     compile "io.micrometer:micrometer-core:$micrometerVersion"
 
     testCompile project(":protobuf")
+
+    testFixturesCompile "junit:junit:$junitVersion"
+    testFixturesCompileOnly "org.projectlombok:lombok:$lombokVersion"
 
     itCompile "junit:junit:$junitVersion"
     itCompile project(":protobuf")
--- a/processor/build.gradle
+++ b/processor/build.gradle
@@ -8,7 +8,6 @@ buildscript {
 }
 
 apply plugin: "me.champeau.gradle.jmh"
-apply plugin: "java-test-fixtures"
 
 dependencies {
     compile project(":common")
--- a/processor/src/main/java/com/linecorp/decaton/processor/TracingProvider.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/TracingProvider.java
@@ -32,6 +32,19 @@ import com.linecorp.decaton.processor.runtime.RetryConfig;
 public interface TracingProvider {
     interface TraceHandle {
         /**
+         * Invoked when processing is considered complete (whether normally or exceptionally).
+         * This will be invoked by the thread that calls {@link DeferredCompletion#complete()},
+         * or a decaton thread if processing was synchronous
+         * (i.e. {@link ProcessingContext#deferCompletion()} was never called).
+         */
+        void processingCompletion();
+    }
+
+    /**
+     * Handle for tracing the execution of a specific {@link DecatonProcessor}
+     */
+    interface ProcessorTraceHandle extends TraceHandle {
+        /**
          * Invoked by the thread that will perform processing immediately before processing
          * Implementations may wish to associate the passed trace with the current thread
          * (e.g. using a {@link ThreadLocal}) for use during {@link DecatonProcessor} execution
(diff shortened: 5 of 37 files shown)


brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTraceHandle.java
```diff
@@ -0,0 +1,60 @@
...
+import brave.Span;
+import brave.messaging.MessagingTracing;
+import brave.propagation.CurrentTraceContext.Scope;
+
+class BraveTraceHandle implements TraceHandle {
+    private final MessagingTracing messagingTracing;
+    private final Span span;
+    private Scope scope;
+
+    public BraveTraceHandle(MessagingTracing messagingTracing, Span span) {
```
kawamuray, line 31: pointless to have a `public` for this since the class itself is package-private.

brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTraceHandle.java
```diff
@@ -0,0 +1,60 @@
...
+        this.messagingTracing = messagingTracing;
+        this.span = span;
+    }
+
+    @Override
+    public void processingStart() {
+        scope = messagingTracing.tracing().currentTraceContext().newScope(span.context());
+        span.annotate("decaton.start");
```
kawamuray, line 39: any reason you renamed this? 
`decaton.start` and `.end` sounds like it covers the whole part of processing inside decaton (so internal queuing etc as well) and confusing I think.

brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTraceHandle.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Override
+    public void processingCompletion() {
+        span.finish();
+    }
+
+    @Override
+    public TraceHandle childFor(DecatonProcessor<?> processor) {
+        final Span childSpan = messagingTracing.tracing().tracer().newChild(span.context())
+                                               .name(processor.getClass().getSimpleName())
```
kawamuray, line 56: Not sure if using a class name always works to identify a particular processor implementation. e.g, if the processor is supplied as a lambda function (which is possible and allowed).
It might be better to add an another method `String name();` to `DecatonProcessor` interface which has default implementation that returns it's class name.

docs/consuming-any-data.adoc
```diff
@@ -1,6 +1,6 @@
 Consuming Arbitrary Topic
 =========================
-:base_version: 0.0.42
+:base_version: 0.0.43
```
kawamuray, line 3: As a process, the person who to make a release will "fix" this version before releasing so please change this to `0.0.43-SNAPSHOT` instead. (overlooked last time)

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessingContextImpl.java
```diff
@@ -90,18 +95,37 @@ public DeferredCompletion deferCompletion() {
...
         }
 
-        CompletableFuture<Void> future = new CompletableFuture<>();
-        DeferredCompletion nextCompletion = () -> future.complete(null);
-
         DecatonTask<P> task = new DecatonTask<>(
                 this.task.metadata(), taskData, this.task.taskDataBytes());
         DecatonProcessor<P> nextProcessor = downstreams.get(0);
+        final TraceHandle parentTrace = request.trace();
+        final TraceHandle traceHandle = null == parentTrace ? NoopTrace.INSTANCE
```
kawamuray, line 102: why is this null-check necessary? I think in all cases a non-null trace is supplied.

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessingContextImpl.java
```diff
@@ -90,18 +95,37 @@ public DeferredCompletion deferCompletion() {
...
+                                                            : parentTrace.childFor(nextProcessor);
+        CompletableFuture<Void> future = new CompletableFuture<>();
+        DeferredCompletion nextCompletion = () -> {
+            future.complete(null);
+            try {
+                traceHandle.processingCompletion();
```
kawamuray, line 108: Please add a test case in `ProcessingContextImpl` for those trace methods invocations.

kawamuray (changes requested):
Thanks for the follow-up :) Overall I agree with this change.

brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTraceHandle.java
m50d: The same method is called in the "parent" case (where it represents the start of the pipeline) and the "child" case (where it represents the start of an individual processor), so I thought `decaton.pipeline.start` was misleading. Maybe just `start` and `return` would be clearer?

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessingContextImpl.java
m50d: The `TracingProvider` could have returned `null`, but a general `try`/`catch` probably covers that better. Will do that.

brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTraceHandle.java
kawamuray: Ah I see. 

I agree with stripping "decaton." prefix because its already scoped by span I think.
However I still think "start" "return" wouldn't be preferred because if I see those annotations in trace the first question would be "what did it started??". So I'd suggest "process.call" "process.return" or something similar that we can tell that it's about process method of processor.

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessingContextImpl.java
kawamuray: I see.
However, as a protocol I think we should consider `childFor`'s return to be always non-null. 
For the guard against misimpementation of driver, I think we should guard against not only NPE but also for other exceptions so `try-catch` should be the way to go.

m50d: Yes, agreed - I've done it that way now.

brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTraceHandle.java
m50d: > So I'd suggest "process.call" "process.return" or something similar that we can tell that it's about process method of processor.

The problem is that it could be either - in the "parent" span it's `pipeline.return` while in the "child" span it's `process.return`. I think it's ok to have the same method / annotation because the span gives it context, but the annotation name needs to be something that works for both cases.

> Besides, is it still meaningful to have .start annotation? I think it's timing is almost the same as its belonging span's creation so it sounds like duplicating the same information.

I agree that the annotation is not very meaningful for processors, but it's important in the "parent" case, so I think it's easier to keep it consistent.

processor/src/test/java/com/linecorp/decaton/processor/runtime/ProcessingContextImplTest.java
```diff
@@ -113,31 +145,60 @@ private static void safeAwait(CountDownLatch latch) {
...
     public void testPush_Level1_Sync() throws InterruptedException {
-        ProcessingContextImpl<HelloTask> context = context((ctx, task) -> { /* noop */ });
-
-        CompletableFuture<Void> future = context.push(TASK);
-        assertTrue(future.isDone());
+        TraceHandle handle = new TestTraceHandle("testPush_Level1_Sync");
+        try {
+            ProcessingContextImpl<HelloTask> context = context(handle,
+                                                               new NamedProcessor("Noop", (ctx, task) ->
+                                                                       assertEquals("testPush_Level1_Sync-Noop",
```
kawamuray, line 152: I think calling asserts inside such closure/callback that is expected to be executed as part of testing target's logic isn't a good pattern because:

* the execution of the closure itself is a part of target's logic and we it doesn't calls it , assertion isn't executed at all
* if the target try-catch the execution and crush the raised exception, assertion won't work properly (not very sure though)

so it's better to have a AtomicReference or something and capture the value inside, test its value in the test method's body.

btw, I think what we want to test for `ProcessingContextImplTest` is that it properly interacts with `TraceHandle` instance, which means beyond the work calling its method like `processStart` is out of scope (it's a responsibility of the driver implementation) of this test case, so this check isn't necessary I think?

processor/src/main/java/com/linecorp/decaton/processor/runtime/TracingProvider.java
```diff
@@ -58,6 +58,8 @@
          * Otherwise, this will be invoked by the thread that calls {@link DeferredCompletion#complete()}.
          */
         void processingCompletion();
+
+        TraceHandle childFor(DecatonProcessor<?> processor);
```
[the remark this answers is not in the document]
kawamuray: btw I should've caught this in last review, but I think this class should be put under `decaton.processor` package rather than `decadton.processor.runtime` as in the former is the place to put public interfaces that's supposed to be implemented by users.

testing/build.gradle
```diff
@@ -18,7 +18,7 @@ ext {
     noPublish = true
 }
 dependencies {
-    implementation project(':processor')
+    implementation testFixtures(project(':processor'))
```
kawamuray, line 21: Even after reading https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures this I still don't get why is this necessary... can you explain what's it for?

kawamuray (changes requested):

processor/src/main/java/com/linecorp/decaton/processor/runtime/TracingProvider.java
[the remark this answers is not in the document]
m50d: > The argument now can be String name? Any extra information you imagine it to obtain from processor instance?

I was thinking some implementations might want to cast / use reflection to access some extra information from the processor instance (e.g. extra tags or metrics).

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessingContextImpl.java
```diff
@@ -90,18 +95,42 @@ public DeferredCompletion deferCompletion() {
...
                 this.task.metadata(), taskData, this.task.taskDataBytes());
         DecatonProcessor<P> nextProcessor = downstreams.get(0);
+        final TraceHandle parentTrace = request.trace();
+        TraceHandle tmpTraceHandle = NoopTrace.INSTANCE;
+        try {
+            tmpTraceHandle = parentTrace.childFor(nextProcessor);
+        } catch(Exception e) {
+            log.error("Exception from tracing", e);
+        }
+        final TraceHandle traceHandle = tmpTraceHandle;
```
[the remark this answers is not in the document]
m50d: Guess so. I feel like this shouldn't be this hard, but I can't see any other way to persuade Java that the value really is `final`.

testing/build.gradle
m50d: I wanted `TestTracingProvider`/`TestTraceHandle` to be used in both `processor` tests and in `testing`, but not to be part of `processor`'s main classpath. So test fixtures seemed like the right place to put them.

processor/build.gradle
```diff
@@ -8,6 +8,7 @@ buildscript {
 }
 
 apply plugin: "me.champeau.gradle.jmh"
+apply plugin: "java-test-fixtures"
```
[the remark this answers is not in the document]
m50d: It creates a source set for test-support code that tests in this project and others can depend on, but that isn't part of the main project. Since there are now tests that use `TestTracingProvider` and `TestTraceHandle` in both `processor` and `testing`.

processor/src/test/java/com/linecorp/decaton/processor/runtime/ProcessingContextImplTest.java
```diff
@@ -47,11 +53,35 @@
...
 import com.linecorp.decaton.protocol.Decaton.DecatonTaskRequest;
 import com.linecorp.decaton.protocol.Decaton.TaskMetadataProto;
 import com.linecorp.decaton.protocol.Sample.HelloTask;
 
 public class ProcessingContextImplTest {
+    private static class NamedProcessor implements DecatonProcessor<HelloTask> {
+        private final String name;
+        private final DecatonProcessor<HelloTask> impl;
+
+        public NamedProcessor(String name, DecatonProcessor<HelloTask> impl) {
```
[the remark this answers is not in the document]
m50d: Seems to be an IntelliJ default. Fixed.

processor/src/test/java/com/linecorp/decaton/processor/runtime/ProcessingContextImplTest.java
m50d: We do need to check that the processor is actually executed by the same thread that called the `TraceHandle` methods one way or another, as that's an important property and fairly easy to break. But agreed with just capturing the value here and then asserting afterwards; will do that.

testing/build.gradle
kawamuray: hm, I see https://stackoverflow.com/questions/5644011/multi-project-test-dependencies-with-gradle/60138176#60138176

processor/src/main/java/com/linecorp/decaton/processor/TracingProvider.java
```diff
@@ -14,14 +14,15 @@
  * under the License.
  */
 
-package com.linecorp.decaton.processor.runtime;
+package com.linecorp.decaton.processor;
 
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 
 import com.linecorp.decaton.client.KafkaProducerSupplier;
 import com.linecorp.decaton.processor.DecatonProcessor;
```
kawamuray, line 22: now those imports become unnecessary.

brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTraceHandle.java
kawamuray: Ah - I see your point.
So what's the purpose of keeping that "parent" code calling `processingStart()` and `processingReturn()`?
`ProcessPipeline#process` calls `context.push(task.taskData());` immediately after calling `processingStart()` so there's no point to keep the call in `ProcessPipeline` I guess?
Then the first (top) processor becomes the direct child of the root span ("decaton") and 2nd processor (if exists) becomes the child of the first processor's span which sounds like exactly what we want.

kawamuray (changes requested):

brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveTraceHandle.java
m50d: At an internal-API level we need `processingStart` and `processingReturn` around the sync processing section to indicate where to bind the span to the current thread (since the thread that fetches from Kafka is not the thread that runs processing). For the "child" case the thread that created the span is also the thread that runs processing, but I think it's clearest to keep the API consistent.

At a zipkin trace level, the `return` annotation is useful to tell us where sync processing finished (I think both for the overall pipeline and for individual processors). But I guess the `start` annotation isn't so useful anymore: we can tell when deserialisation etc. finished by looking at when the child span started.

I'll remove the `start` annotation and rename `decaton.return` -> `return`?

kawamuray: Hmm, I don't understand.

> At an internal-API level we need processingStart and processingReturn around the sync processing section to indicate where to bind the span to the current thread (since the thread that fetches from Kafka is not the thread that runs processing). For the "child" case the thread that created the span is also the thread that runs processing, but I think it's clearest to keep the API consistent.

Even if we delete `processStart` and `processReturn` at `ProcessorPipeline`, the `processStart` call at `ProcessingContextImpl` just works as the point to attach the context to the current thread right?
Because what tracing drivers wants to measure separately from "decaton's internal processing" is the duration of each processor's processing time as you said, when we consider the optimal interface, there should be methods that are called just once before a processor starts processing and just once after a processor returns from processing I think, and I don't understand how having another point of call them at `ProcessorPipeline` could be "consistent" from API's point of view. Rather it feels like reusing a method prepared for another purpose.

> (I think both for the overall pipeline and for individual processors).

How useful it is while two annotations (one from pipeline, one from the top processor) points almost exactly at the same point?

To confirm my understanding, with your current implementation the graphs will be like follows right?
The first `.start` `.return` annotations from pipeline and the another pair from the 1st processor shows almost the same point (there's few instructions between the points but that should be negligible and actually invisible).
```
[--                                                            "decaton" span                                                --]
                                                       ^.start                                                                ^.return // from ProcessorPipeline
<- decaton internal moment->     [--                   top processor's span                         --]
                                                       ^start                                                                 ^return  // from the top processor
                                                                            [--                   2nd processor's span   --]
                                                                            ^start                                            ^return
```

>  But I guess the start annotation isn't so useful anymore: we can tell when deserialisation etc. finished by looking at when the child span started.

Agree with this.

m50d: > Even if we delete processStart and processReturn at ProcessorPipeline, the processStart call at ProcessingContextImpl just works as the point to attach the context to the current thread right?
> Because what tracing drivers wants to measure separately from "decaton's internal processing" is the duration of each processor's processing time as you said, when we consider the optimal interface, there should be methods that are called just once before a processor starts processing and just once after a processor returns from processing I think, and I don't understand how having another point of call them at ProcessorPipeline could be "consistent" from API's point of view. Rather it feels like reusing a method prepared for another purpose.

Hmm. In terms of how the brave API is intended to be used, spans are meant to be bound to threads as much as possible. But concretely I guess most of the execution of `ProcessPipeline` is covered by the child processors and so binding the parent span is redundant. Fair enough.

In that case I think a full split between parent/child span interfaces makes sense.

> How useful it is while two annotations (one from pipeline, one from the top processor) points almost exactly at the same point?

In the UI we're currently using it's not possible to view annotations from more than one span at the same time. So it's good if each span has whatever annotations are needed to understand that span's behaviour without having to switch back and forth (indeed I'm second-guessing removing the annotation that's at the same place as where the child span starts, since that's not actually so visible when looking at the details of the parent span). But I guess that's a limitation of this specific UI and maybe decaton can't be expected to work around it.

kawamuray: >  I guess most of the execution of ProcessPipeline is covered by the child processors

If you mean few lines of code like  by "a few of execution" (in contrast to "most of the execution")) here, it should be totally negligible and pointless to measure as I wrote: 

> (there's few instructions between the points but that should be negligible and actually invisible)

> In the UI we're currently using it's not possible to view annotations from more than one span at the same time. So it's good if each span has whatever annotations are needed to understand that span's behaviour without having to switch back and forth 

Yeah, but the thing is, the ".start" annotation marked on the root span at `ProcessorPipeline` and the starting point of the child span created for the 1st processor points exactly the same time (there can be few microseconds or nanoseconds between, because there are few jvm instructions indeed) and so I think we don't even have to refer the annotation to see the starting point of the processor pipeline. For the `.return` annotation that might not be the case, but the issue sounds trivial enough.

m50d: > Yeah, but the thing is, the ".start" annotation marked on the root span at ProcessorPipeline and the starting point of the child span created for the 1st processor points exactly the same time (there can be few microseconds or nanoseconds between, because there are few jvm instructions indeed) and so I think we don't even have to refer the annotation to see the starting point of the processor pipeline.

Right, but when looking at the parent span it's actually quite non-obvious where the starting point of the child span is. (And annotations on the child span are not visible at all without clicking back and forth). I agree that it's trivial on a technical level, but it's actually quite a serious difference in usability when a lot of the value of these tools comes from being able to get a visual overview of where execution time is being spent without having to calculate any numbers or crossreference anything.

But I don't think that's decaton's responsibility, and I've followed your approach with the most recent commit - please take a look and see what you think.

brave/src/main/java/com/linecorp/decaton/processor/runtime/BraveProcessorTraceHandle.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Override
+    public void processingStart() {
+        scope = messagingTracing.tracing().currentTraceContext().newScope(span.context());
+    }
+
+    @Override
+    public void processingReturn() {
+        span.annotate("return");
+        scope.close();
```
kawamuray, line 40: Note for myself and possible future reviewers:

This is for setting span into the current thread and then revert it back to "unset" (or previously set scope) and is not for closing the span (`span.finish()`).

kawamuray (approved):
LGTM, thanks for the follow-up :)

kawamuray:
This change has been released as part of decaton 0.0.43.

PR #72, opened 2020-10-23
Add interface to decatonClient for customizing delay

4Jessie:
Hello, decaton team.
We are using decaton in our application and for one topic, we’d like to use the delayed event feature of decaton.  We are using decatonClient to publish events and there is no put method which can be used to custom  scheduledTime  and it’s hard to extend the DecatonClientImpl class. So I added the put method with delay duration (by duration or milliseconds) as parameter. 
Please take a look at it.

### DIFF
--- a/client/src/main/java/com/linecorp/decaton/client/DecatonClient.java
+++ b/client/src/main/java/com/linecorp/decaton/client/DecatonClient.java
@@ -16,11 +16,13 @@
 
 package com.linecorp.decaton.client;
 
-import java.time.Duration;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
 
 import com.linecorp.decaton.common.Serializer;
+
+import lombok.Builder;
+import lombok.Value;
 
 /**
  * Decaton client interface to use for putting tasks on decaton queue.
@@ -51,20 +53,10 @@ public interface DecatonClient<T> extends AutoCloseable {
      * Put a task onto associated decaton queue with specifying arbitrary delay duration.`
      * @param key the criteria to shuffle and order tasks. null can be specified if it doesn't matters.
      * @param task an instance of task. Should never be null.
-     * @param delayDuration duration which is used to set delay of task metadata.
+     * @param overrideTaskMetadata taskMetaData which can be set by users and used for event publish.
      * @return a {@link CompletableFuture} which represents the result of task put.
      */
-    CompletableFuture<PutTaskResult> put(String key, T task, Duration delayDuration);
-
-    /**
-     * Put a task onto associated decaton queue with specifying arbitrary timestamp and arbitrary delay.
-     * @param key the criteria to shuffle and order tasks. null can be specified if it doesn't matters.
-     * @param task an instance of task. Should never be null.
-     * @param timestamp milliseconds precision timestamp which is to be used to set timestamp of task metadata.
-     * @param delayInMillis milliseconds which is used to set delay duration of task metadata.
-     * @return a {@link CompletableFuture} which represents the result of task put.
-     */
-    CompletableFuture<PutTaskResult> put(String key, T task, long timestamp, long delayInMillis);
+    CompletableFuture<PutTaskResult> put(String key, T task, TaskMetaData overrideTaskMetadata);
 
     /**
      * Put a task onto associated decaton queue.
@@ -131,5 +123,14 @@ public interface DecatonClient<T> extends AutoCloseable {
      */
     static <T> DecatonClientBuilder<T> producing(String topic, Serializer<T> serializer) {
         return new DecatonClientBuilder<>(topic, serializer);
+    }
+
+    @Builder
+    @Value
+    class TaskMetaData {
+        /** timestamp of task metadata */
+        Long timestamp;
+        /** scheduledTime for event processing, it should be (timestamp + delayDuration) in milliseconds */
+        Long scheduledTime;
     }
 }
--- a/client/src/main/java/com/linecorp/decaton/client/DecatonClient.java
+++ b/client/src/main/java/com/linecorp/decaton/client/DecatonClient.java
@@ -16,6 +16,7 @@
 
 package com.linecorp.decaton.client;
 
+import java.time.Duration;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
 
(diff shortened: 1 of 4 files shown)


client/src/main/java/com/linecorp/decaton/client/DecatonClient.java
```diff
@@ -46,6 +47,25 @@
...
+
+    /**
+     * Put a task onto associated decaton queue with specifying arbitrary timestamp and arbitrary delay.
+     * @param key the criteria to shuffle and order tasks. null can be specified if it doesn't matters.
+     * @param task an instance of task. Should never be null.
+     * @param timestamp milliseconds precision timestamp which is to be used to set timestamp of task metadata.
+     * @param delayInMillis milliseconds which is used to set delay duration of task metadata.
+     * @return a {@link CompletableFuture} which represents the result of task put.
+     */
+    CompletableFuture<PutTaskResult> put(String key, T task, long timestamp, long delayInMillis);
```
kawamuray, line 67: Thanks for the PR! 
I know that there are several people who're building a task metadata manually and produce to enable scheduled execution (which we added just for retry but unexpectedly leveraged for other purpose) so I think this change will be helpful for many users 👍 

Just one concern is, if we keep adding more overloads of `put` methods along with the list of metadata properties to support that'll make the interface to have massive amount of overloads that combines many different sets of metadata properties to customize.
So at this point I think we should introduce a interface like:
```java
     @Value
     @Builder
     class TaskMetadata {
         Long timestamp;
         Long executeTime;
     }
     CompletableFuture<PutTaskResult> put(String key, T task, TaskMetadata overrideMetadata); 
```

and let users to use it in the way:
```java
decatonClient.put(..., TaskMetadata.builder().setScheduledTime(1234).build());
```

WDYT?

kawamuray (changes requested):

client/src/main/java/com/linecorp/decaton/client/DecatonClientImpl.java
```diff
@@ -77,6 +78,31 @@ public DecatonClientImpl(String topic,
         return producer.sendRequest(key, request);
     }
 
+    @Override
+    public CompletableFuture<PutTaskResult> put(String key, T task, Duration delayDuration) {
+        return put(key, task, timestampSupplier.get(), delayDuration.toMillis());
+    }
+
+    @Override
+    public CompletableFuture<PutTaskResult> put(String key, T task, long timestamp, long delayInMillis) {
```
mauhiz, line 87: Just a detail, I think `delay` should be a `Duration` here too

mauhiz (approved):

client/src/main/java/com/linecorp/decaton/client/DecatonClient.java
4Jessie: Yes, that's better.  I was planning to do that when adding the new methods,  but since there are already some put methods defined with the timeStamp, finally I decided to add methods only with scheduledTime.  In view of the possible usage and possible modification on taskMetaData in the future, your recommanded way is much better (although that kind of definition is also able to be used to provide same function as the existing methods with timestamp as parameter).

client/src/main/java/com/linecorp/decaton/client/DecatonClient.java
```diff
@@ -46,6 +49,15 @@
      */
     CompletableFuture<PutTaskResult> put(String key, T task, long timestamp);
 
+    /**
+     * Put a task onto associated decaton queue with specifying arbitrary delay duration.`
```
kawamuray, line 53: "specifying arbitrary delay duration." => "specifying some fields of task metadata."

client/src/main/java/com/linecorp/decaton/client/DecatonClient.java
```diff
@@ -112,4 +124,13 @@
     static <T> DecatonClientBuilder<T> producing(String topic, Serializer<T> serializer) {
         return new DecatonClientBuilder<>(topic, serializer);
     }
+
+    @Builder
+    @Value
+    class TaskMetaData {
```
kawamuray, line 130: I think we name it `Metadata` rather than `MetaData` conventionally.  So `TaskMetadata` would be better.

client/src/main/java/com/linecorp/decaton/client/DecatonClient.java
```diff
@@ -112,4 +124,13 @@
...
+        /** timestamp of task metadata */
```
kawamuray, line 131: let's make it a valid javadoc :)

```java
/**
 * Timestamp of task metadata.
 */
```

client/src/main/java/com/linecorp/decaton/client/DecatonClientImpl.java
```diff
@@ -77,6 +77,19 @@ public DecatonClientImpl(String topic,
...
+    public CompletableFuture<PutTaskResult> put(String key, T task, TaskMetaData overrideTaskMetaData) {
+        byte[] serializedTask = serializer.serialize(task);
+
+        DecatonTaskRequest request =
+                DecatonTaskRequest.newBuilder()
+                                  .setMetadata(convertToTaskMetadataProto(overrideTaskMetaData))
+                                  .setSerializedTask(ByteString.copyFrom(serializedTask))
+                                  .build();
+
+        return producer.sendRequest(key, request);
```
kawamuray, line 90: let's add a private method like below and use it also from here https://github.com/line/decaton/pull/72/files#diff-b62f1c35402d8557a005fbb36c449954296e0efcb21d473196e04029d4c6fc33R65 to avoid duplicating common parts.

```java
private CompletableFuture<PutTaskResult> put(String key, T task, TaskMetadataProto metadataProto)
```

client/src/main/java/com/linecorp/decaton/client/DecatonClientImpl.java
```diff
@@ -91,4 +104,29 @@ public DecatonClientImpl(String topic,
     public void close() throws Exception {
         producer.close();
     }
+
+    private TaskMetadataProto convertToTaskMetadataProto(TaskMetaData overrideTaskMetadata) {
+        final TaskMetadataProto.Builder
+                taskMetadataProtoBuilder = TaskMetadataProto.newBuilder()
+                                                            .setSourceApplicationId(applicationId)
+                                                            .setSourceInstanceId(instanceId);
+        if (timestampSupplier != null) {
```
kawamuray, line 113: `timestampSupplier` is configurable only through package-private constructor and in public constructor, it is supplied by default, so you don't have to expect it to be null.

client/src/main/java/com/linecorp/decaton/client/DecatonClientImpl.java
```diff
@@ -91,4 +104,29 @@ public DecatonClientImpl(String topic,
...
+            taskMetadataProtoBuilder.setTimestampMillis(timestampSupplier.get());
+        } else {
+            taskMetadataProtoBuilder.setTimestampMillis(System.currentTimeMillis());
+        }
+        if (overrideTaskMetadata == null) {
```
kawamuray, line 118: I think this is never be a case, so we don't need to check null for it either.

client/src/main/java/com/linecorp/decaton/client/DecatonClientImpl.java
```diff
@@ -91,4 +104,29 @@ public DecatonClientImpl(String topic,
...
+            return taskMetadataProtoBuilder.build();
+        }
+        final Long timestamp = overrideTaskMetadata.getTimestamp();
+        final Long scheduledTime = overrideTaskMetadata.getScheduledTime();
+        if (timestamp != null && timestamp > 0) {
```
kawamuray, line 123: If we juedge negative timestamp to be invalid here, I think we should throw `IllegalArgumentException` for it rather than silently dropping the value.
But I think we can leave the value of it as the responsibility of users, so we may just check if it's null or not here.

client/src/main/java/com/linecorp/decaton/client/DecatonClientImpl.java
```diff
@@ -91,4 +104,29 @@ public DecatonClientImpl(String topic,
...
+            taskMetadataProtoBuilder.setTimestampMillis(timestamp);
+        }
+        if (scheduledTime != null && scheduledTime > 0) {
```
kawamuray, line 126: Same comment applies here.

kawamuray (changes requested):
Just few minor suggestions.

kawamuray (approved):
LGTM 👍 Thanks for your contribution!

kawamuray:
Build failure by doc version tag. I'll take care of it after merging.

kawamuray:
This change has been released as part of [0.0.44](https://github.com/line/decaton/releases/tag/v0.0.44)

PR #74, opened 2020-11-20
Make record header accessible from DecatonProcessor fixes #73

### LINKED ISSUE #73: Add kafka headers support
Hi, decaton team.
I suggest the basic function of Kafka.
Decaton does not support consumer header information. If decaton provides header information, it will be useful to all users.
For example, I need 3 Kafka headers(`kafka_replyTopic`, `kafka_replyPartition`, `kafka_correlationId`) to use the request-reply pattern:))

ocadaruma:
## Motivation
- Currently there's no way to refer Kafka's RecordHeader from DecatonProcessor
- Sometimes it's necessary to access metadata in the headers and propagate it from DecatonProcessor
  * e.g. Spring-Kafka's request/reply semantics uses KafkaHeader to control message routing

## Changes
- This PR makes DecatonProcessor to access Headers through `ProcessingContext`

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessingContextImpl.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessingContextImpl.java
@@ -39,7 +39,6 @@ public class ProcessingContextImpl<T> implements ProcessingContext<T> {
     private final String subscriptionId;
     private final TaskRequest request;
     private final DecatonTask<T> task;
-    private final Headers headers;
     private final DeferredCompletion completion;
     private final List<DecatonProcessor<T>> downstreams;
     private final DecatonProcessor<byte[]> retryQueueingProcessor;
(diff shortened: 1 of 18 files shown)


processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessingContextImpl.java
```diff
@@ -73,6 +77,11 @@ public String key() {
         return request.key();
     }
 
+    @Override
+    public Headers headers() {
+        return headers;
```
kawamuray, line 82: Any points to add a `field`? Like other getters just returning `request.headers()`  works?

kawamuray (changes requested):
Just one minor point.

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessingContextImpl.java
ocadaruma: Oh nice catch, completely right.

kawamuray (approved):
LGTM, nice 👍

PR #77, opened 2020-12-15
Add document explaining new versioning policy that is based on semantic versioning

kawamuray:


VERSIONING.md
```diff
@@ -0,0 +1,58 @@
...
+You can see the list of changes that has been made by particular version at [Releases page](https://github.com/line/decaton/releases).
+
+
+Guide for developers
+====================
+
+It is a developer's (and reviewer's) responsibility to properly bump the required part of the version number along with the patch for the intended changes.
+Upon submitting a PR, one should include the diff for changing the version according to the content of the patch.
+For example, if the current version is `1.2.0`, if the PR contains a newly added public API that doesn't break backward compatibility, the PR should bump the version to `1.3.0`.
+If there is a conflict with changing the version upon merging PR, the greater version should always be adopted. For example, if the previously merged PR bumped the version to `1.3.0` but the PR now being merged has set to `1.2.1`, then the author of the PR being merged should be rebased onto the latest master to ensure the version is fixed at `1.3.0`.
```
ocadaruma, line 58: So the version of the PR now being merged should be `1.3.1` right?

kawamuray: No. 

* The PR that has already merged and bumped the version to `1.3.0` contained new feature
* The PR now being merged contains only a compatible fix

A minor version bump covers and changes that requires patch version to get bumped, and major version works the same for minor and patch. So when the version has already bumped to `1.3.0` which has larger minor version than the last release, any patch level changes does not need to touch patch version.

ocadaruma: Understood, thanks for your explanation.

kawamuray:
@ocadaruma I updated the doc and build configuration based on our discussion.

In new flow, we will stop bumping upcoming release version and document versions. 
Instead, each PRs are marked with labels and the release handler will take care of bumping version properly.

ocadaruma (approved):
Thanks, LGTM.
Let's try new semver-based workflow.

PR #81, opened 2021-01-14
Use auto.offset.reset=earliest for benchmark

kawamuray:
by #60 , an invocation of `KafkaConsumer#position` has been deleted from `ConsumerRebalanceListener#onPartitionsAssigned`. This seems to have caused a tiny behavior change, in particular, the initial offset reset now happens in later timing than before, which was occurring at calling `position()` in previous implementation.
This led our benchmark to break by timing out awaiting all produced records to have completed, because in the current implementation, the offset hasn't yet reset to 0 when the benchmark program starts producing tasks, causing a consumer in `ProcessorSubscription` to reset the offset to some value larger than 0 because there's already some records present in the topic, and the offset reset policy is `latest` (default).

Thanks a lot for @ocadaruma for investigating this!


ocadaruma (approved):
Thanks for quick fix. LGTM !

PR #82, opened 2021-01-15
Log error trace when scheduleThenProcess fails

mauhiz:
We found this cryptic error in our project (".. Offset of this will left uncommitted") but could not tell the cause since there is no stacktrace and no other log.

In this PR, I separate exception handling into
- InterruptedException : more or less expectable on shutdown; log at warn level
- the rest : really unexpected since `ProcessPipeline` also does its share of error handling

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorUnit.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorUnit.java
@@ -66,11 +66,13 @@ public class ProcessorUnit implements AsyncShutdownable {
         Timer timer = Utils.timer();
         try {
             pipeline.scheduleThenProcess(request);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            log.warn("Scheduling and/or processing interrupted. Shutdown initiated={}, request: {}",
+                     terminated, request);
         } catch (Exception e) {
-            if (e instanceof InterruptedException) {
-                Thread.currentThread().interrupt();
-            }
-            log.error("Error while processing request {}. Offset of this will left uncommitted", request);
+            log.error("Error while processing request {}. Corresponding offset will be left uncommitted.",
+                      request, e);
         } finally {
             // This metric measures the total amount of time the processors were processing tasks including time
             // for scheduling those tasks and is used to refer processor threads utilization, so it needs to measure
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorUnit.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorUnit.java
@@ -70,7 +70,8 @@ public class ProcessorUnit implements AsyncShutdownable {
             if (e instanceof InterruptedException) {
                 Thread.currentThread().interrupt();
             }
-            log.error("Error while processing request {}. Offset of this will left uncommitted", request);
+            log.error("Error while processing request {}. Corresponding offset will be left uncommitted.",
+                      request, e);
         } finally {
             // This metric measures the total amount of time the processors were processing tasks including time
             // for scheduling those tasks and is used to refer processor threads utilization, so it needs to measure
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorUnit.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorUnit.java
@@ -66,13 +66,11 @@ public class ProcessorUnit implements AsyncShutdownable {
         Timer timer = Utils.timer();
         try {
             pipeline.scheduleThenProcess(request);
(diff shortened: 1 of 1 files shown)


processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorUnit.java
```diff
@@ -66,11 +66,13 @@ private void processTask(TaskRequest request) {
         Timer timer = Utils.timer();
         try {
             pipeline.scheduleThenProcess(request);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            log.warn("Scheduling and/or processing interrupted. Shutdown initiated={}, request: {}",
+                     terminated, request);
```
ocadaruma, line 72: Isn't better to pass `e` ?

ocadaruma (commented):
Thank you for the patch.
Left minor comment.

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorUnit.java
```diff
@@ -66,11 +66,13 @@ private void processTask(TaskRequest request) {
...
         } catch (Exception e) {
-            if (e instanceof InterruptedException) {
-                Thread.currentThread().interrupt();
-            }
-            log.error("Error while processing request {}. Offset of this will left uncommitted", request);
```
kawamuray, line 73: First of all, I admit this wasn't a good log message ... :p 

This logging was added by #32, which I attempted to improve code around calling user-supplied DecatonProcessors.
This place was originally handling any exceptions including those from decaton core and those from user-supplied `DecatonProcessor`, but after the patch, now this place is supposed to catch exceptions only from decaton core since exceptions from `DecatonProcessor` is handled here: https://github.com/line/decaton/blob/master/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessPipeline.java#L113

I also dropped special handling of `InterruptedException` by the reason I explained in PR comment:
> After this patch the existing special treatment against InterruptedException during shutdown is dropped because we found that it is pointless in typical implementation o processors which wraps entire process() logic with wide try-catch and handling in user's code. The reason it was only for InterruptedException is that it is one of the most typically thrown exception during shutdown but it lacks strong rationale since any other exceptions may be thrown instead of it (especially or libraries that prefer unchecked exceptions).

which should still be true. 

So if you got this error message by an `InterruptedException`, I think that means its from here: https://github.com/line/decaton/blob/master/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessPipeline.java#L77 , but what I don't understand now is, despite it attempts to gracefully return from there by closing `CountDownLatch` corresponding to the call of `close()`, why (and who) interrupted into the thread of your decaton application.

How do you configure shutdown sequence of your application, and is it normal that the decaton thread gets interrupted quite immediately (even before close?) on shutdown?

The intention of this log message was to warn users that the handling of this record is out of decaton's control (due to an unexpected exception) and they need to confirm that it happend during irregular time like at shutdown, so I thought it should be there even if the exception is `InterruptedException`.

kawamuray (changes requested):

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorUnit.java
mauhiz: I'm not sure why we'd want to log what sleep was interrupted - does it help?

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorUnit.java
mauhiz: We didn't get this exception on shutdown but during regular processing; and I have no reason to think it's an `InterruptedException`.

> I also dropped special handling of `InterruptedException`

Well not completely :p and having 2 catch branches is not significantly worse than an instanceof check.

The only thing I really want to do in this PR is log the exception trace; so that I can get more clues. If you prefer, I can revert all but adding the trace?

kawamuray: > We didn't get this exception on shutdown but during regular processing; and I have no reason to think it's an InterruptedException.

Ah I see, misunderstand that.

> Well not completely :p and having 2 catch branches is not significantly worse than an instanceof check.

yup.

> The only thing I really want to do in this PR is log the exception trace; so that I can get more clues. If you prefer, I can revert all but adding the trace?

Yeah, let's take that step first. I'd like to know why and how the exception raised at that path (which I didn't expected to happen in typical application during typical shutdown sequence) , so we can consider right workaround after knowing what happened in your application first.

kawamuray (approved):

kawamuray:
LGTM, thanks!

PR #83, opened 2021-01-20
Separate internal-only APIs from public API by package

kawamuray:
As a preparation for starting semantic versioning.
Basic strategy:
* Introduce `internal` subpackage in places we need to have internal-only package
* Trivial classes may use package-private visibility in the same package.
* Public API is defined as classes either NOT 1. set to non-public visibility or 2. classes under the `internal` subpackage
* Decided to not fully relying on Java visibility because of several reasons:
  * Doesn't support JAR level visibility (only package-private)
  * Some classes may left accessible from advanced users. We don't guarantee it's future compatibility, but sometimes it's useful for users who knows internal structure well to have access for internal package for testing/hotfix patching.

Also organized package structure by moving classes and introducing some new packages.
* root package => interfaces/classes to implement `DecatonProcessor`
* `runtime` => interface/classes to instantiate a decaton runtime (`ProcessorSubscription`)
* `processors` => helper processor implementations. `CompactionProcessor` so far. 
* `tracing` => tracing related.

I'll update VERSIONING document in separate PR.


processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ProcessingContextImpl.java
```diff
@@ -14,7 +14,7 @@
  * under the License.
  */
 
-package com.linecorp.decaton.processor.runtime;
+package com.linecorp.decaton.processor.runtime.internal;
```
ocadaruma, line 17: Shouldn't this class be placed in root/internal ? (i.e. `com.linecorp.decaton.processor.internal`)
As `ProcessingContext<T>` is in root.

processor/src/main/java/com/linecorp/decaton/processor/runtime/TaskMetadata.java
```diff
@@ -14,7 +14,7 @@
  * under the License.
  */
 
-package com.linecorp.decaton.processor;
+package com.linecorp.decaton.processor.runtime;
 
 import com.linecorp.decaton.protocol.Decaton.TaskMetadataProto;
```
ocadaruma, line 19: I think this class should be placed in root because we use it when implementing DecatonProcessor to retrieve `retryCount`

processor/src/main/java/com/linecorp/decaton/processor/runtime/LoggingContext.java
```diff
@@ -19,7 +19,7 @@
 import org.slf4j.MDC;
```
ocadaruma, line 19: I think this class should be placed in root because we use it when implementing DecatonProcessor to output logs with MDC infos.

ocadaruma (commented):
Thanks for the work!

Besides of the comments, let me confirm about the definition of "Public API".

- If a change about `internal` classes causes a change in `public` classes, it could be considered as "breaking change" right?
  * e.g. If we rename `TaskRequest` to `InternalTaskRequest` (or something), it causes breaking change in `LoggingContext`'s constructor signature

processor/src/main/java/com/linecorp/decaton/processor/runtime/TaskMetadata.java
kawamuray: oops nice catch! thanks.

processor/src/main/java/com/linecorp/decaton/processor/runtime/LoggingContext.java
kawamuray: This too, nice catch.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ProcessingContextImpl.java
kawamuray: Exactly. Thanks for pointing out.

kawamuray:
> If a change about internal classes causes a change in public classes, it could be considered as "breaking change" right? 

no. `public LoggingContext(TaskMetadata)` receives an instance of `TaskRequest`, which is in internal package. So in order to use that API, users have to touch a private API (TaskRequest), which we do not expect. So despite the method (constructor) is at public class, since we don't expect any users are now using it (under assumption of its future compatibility I mean), we do not consider it as a public API and hence isn't a breaking change.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ProcessingContextImpl.java
kawamuray: Resindered, I think no, because while the interface doesn't assume any parts from the runtime package, the Impl requires instances that are of `runtime` package and can't be constructed without that.

ocadaruma: I see, understood.

ocadaruma (approved):
Thanks for the explanation, got it.
> we do not consider it as a public API and hence isn't a breaking change.

Then LGTM !

PR #84, opened 2021-01-22
Update VERSIONING doc following the recent public API separation

kawamuray:
ref: #83

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -19,7 +19,7 @@ We consider the following items as Public API and each one of the changes made a
 
 * Public interfaces/classes/enums that are defined by source code in this project and are included in a published JAR, excluding ones with the following conditions.
   * An interface/class/enum that belongs to a package containing subpackage `internal`. e.g, `com.linecorp.decaton.runtime.internal.Aclass`.
-  * A method/constant/variable that depends on at least one of internal items in its signature. e.g, `com.linecorp.decaton.LoggingContext`'s constructor that takes `com.linecorp.decaton.runtime.internal.TaskRequest` as an argument.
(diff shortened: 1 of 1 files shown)


VERSIONING.md
```diff
@@ -14,11 +14,12 @@ Given a version number MAJOR.MINOR.PATCH, increment the:
...
+We consider the following items as Public API and each one of the changes made against those items in a backward-incompatible way are considered a "breaking change", so we bump the MAJOR version for the release shipping the change.
 
 
-* Public interfaces, classes, and enums that are defined by source code in this project and are included in a published JAR
-  * NOTE: We will likely separate modules for public-APIs and non-public-APIs for the `processor` package in future
+* Public interfaces/classes/enums that are defined by source code in this project and are included in a published JAR, excluding ones with the following conditions.
+  * An interface/class/enum that belongs to a package containing subpackage `internal`. e.g, `com.linecorp.decaton.runtime.internal.Aclass`.
+  * A method/constant/variable that depends on at least one of internal items in its signature. e.g, `com.linecorp.decaton.LoggingContext`'s constructor that takes `com.linecorp.decaton.runtime.internal.TaskRequest` as an argument.
```
ocadaruma, line 22: Just `constructor/method/field` would be more precise?

- `variable` sounds like ambiguous between field/local
  * And we don't care about local variable, of course
- then if we define include `field` in the list, `constant` sounds redundant because in java, constant is `static final` field
- I think "signature" doesn't make sense for fields, so should be mentioned separately? like `public field that has a type of internal item`

I know this is kind of nitpicking though

kawamuray: That's a good point. I'll fix it.

ocadaruma (approved):
LGTM, thanks for update

PR #85, opened 2021-01-25
Update release script to follow the new versioning scheme

kawamuray:
As we're going to manage versions following semver we're gonna need to specify the release version by ourselves.
Also deprecating stuff related to SNAPSHOT versions as we very rarely make a publish with a SNAPSHOT version.


ocadaruma (approved):
LGTM

PR #86, opened 2021-01-27
Fix regression handling

kawamuray:
Fix a bug introduced by #60 .

Handling of `OffsetRegressionException` calls `assignManager.repair` which internally replaces the `PartitionContext` corresponding to the topic-partition, but it later still uses an old, previously grabbed `Context` causing subsequent `registerOffset` to never succeeds.


ocadaruma (approved):
Thanks for the fix. LGTM

PR #87, opened 2021-01-28
DocChecker: Fail validation only if base_version differs to the release version by major version

kawamuray:
Background:
* We have started semantic versioning
* DocChecker always makes a release build to fail because most of the docs depends on `processor` module which gets updated almost 100% on each release.

Let's change DocChecker to fail only if `:base_version:` has lower major version than the version being released. 
This should be okay because:
* The objective of DocChecker is to avoid a document being outdated by forgotten to be updated
* A document gets outdated mainly by Public API changes. Otherwise, the content in doc is almost guaranteed to keep working.


ocadaruma (approved):
LGTM

PR #88, opened 2021-02-08
Bugfix: Exception from consumer.pause() in rebalance due to passing already revoked partition

kawamuray:
In current implementation the `consumer.pause()` may throw `IllegalStateException: No current assignment for partition ...`. 

When rebalance occurs, we save the list of partitions and restore the pause state later. However, since the rebalance might have taken over some partitions off from this instance, we have to filter these partitions from the argument of consumer.pause().
So this happens when all the following 3 conditions are met:
* There’s a paused partition
* Trigger rebalance (instance failure, restart)
* Rebalance taken over a partition from the current instance


ocadaruma (approved):
LGTM, Thanks for quick fix

PR #90, opened 2021-02-17
Migrate to GitHub Actions

ocadaruma:
## Summary
- By renewal of Travis CI's pricing plan last year, we'll likely to hit the build limit
  * https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
- Many OSS are moving to GitHub Actions. I think it's time to move for Decaton as well


.github/workflows/ci.yml
```diff
@@ -0,0 +1,19 @@
+name: CI
+on: [pull_request, push]
```
kawamuray, line 2: Would it be better to limit the target branches as explained in: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-multiple-events-with-activity-types-or-configuration ?

kawamuray (changes requested):

.github/workflows/ci.yml
ocadaruma: Ah right, `pull_request` should be limited to `master`

ocadaruma: Ah no, specifying `master` for both should be fine?

kawamuray:
LGTM, thanks for quick handling!

kawamuray (approved):

PR #93, opened 2021-02-25
Add integration tests for BraveTracingProvider

m50d:
We already have an integration test for tracing in general, but it's good to confirm that the specific integration with Brave also works

### DIFF
--- a/brave/src/it/java/com/linecorp/decaton/processor/runtime/BraveTracingTest.java
+++ b/brave/src/it/java/com/linecorp/decaton/processor/runtime/BraveTracingTest.java
@@ -1,0 +1,119 @@
+/*
+ * Copyright 2020 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.decaton.processor.runtime;
+
+import static org.junit.Assert.assertEquals;
+
+import java.time.Duration;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ThreadLocalRandom;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import com.linecorp.decaton.client.DecatonClientBuilder.DefaultKafkaProducerSupplier;
+import com.linecorp.decaton.processor.TaskMetadata;
+import com.linecorp.decaton.testing.KafkaClusterRule;
+import com.linecorp.decaton.testing.TestUtils;
+import com.linecorp.decaton.testing.processor.ProcessedRecord;
+import com.linecorp.decaton.testing.processor.ProcessingGuarantee;
+import com.linecorp.decaton.testing.processor.ProcessingGuarantee.GuaranteeType;
+import com.linecorp.decaton.testing.processor.ProcessorTestSuite;
+import com.linecorp.decaton.testing.processor.ProducedRecord;
+
+import brave.Tracing;
+import brave.kafka.clients.KafkaTracing;
+
+public class BraveTracingTest {
+
+    private KafkaTracing kafkaTracing;
+    private Tracing tracing;
+
+    public static class BraveTracePropagation implements ProcessingGuarantee {
+        private final Map<String, String> producedTraceIds = new ConcurrentHashMap<>();
+        private final Map<String, String> consumedTraceIds = new ConcurrentHashMap<>();
+        private final Tracing tracing;
+
+        public BraveTracePropagation(Tracing tracing) {this.tracing = tracing;}
+
+        @Override
+        public void onProduce(ProducedRecord record) {
+            producedTraceIds.put(record.task().getId(), tracing.currentTraceContext().get().traceIdString());
+        }
+
+        @Override
+        public void onProcess(TaskMetadata metadata, ProcessedRecord record) {
+            consumedTraceIds.put(record.task().getId(), tracing.currentTraceContext().get().traceIdString());
+        }
+
+        @Override
+        public void doAssert() {
+            assertEquals(producedTraceIds, consumedTraceIds);
+        }
+    }
+
+    @ClassRule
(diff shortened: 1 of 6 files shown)


brave/src/it/java/com/linecorp/decaton/processor/runtime/BraveTracingTest.java
```diff
@@ -0,0 +1,151 @@
...
+                .excludeSemantics(
+                        GuaranteeType.PROCESS_ORDERING,
+                        GuaranteeType.SERIAL_PROCESSING)
+                .customSemantics(new BraveTracePropagation(tracing))
+                .build()
+                .run();
+    }
+
+    @Test(timeout = 30000)
+    public void testTracePropagationWithNullKey() throws Exception {
```
ocadaruma, line 121: Trace id is propagated through header, so whether the key is null or not doesn't affect I think?
Do you have any concrete case that this test will detect a problem?

brave/src/it/java/com/linecorp/decaton/processor/runtime/BraveTracingTest.java
```diff
@@ -0,0 +1,151 @@
...
+        final DefaultKafkaProducerSupplier producerSupplier = new DefaultKafkaProducerSupplier();
+        ProcessorTestSuite
+                .builder(rule)
+                .configureProcessorsBuilder(builder -> builder.thenProcess((ctx, task) -> {
+                    if (ctx.metadata().retryCount() == 0 && ThreadLocalRandom.current().nextBoolean()) {
+                        ctx.retry();
+                    }
+                }))
+                .producerSupplier(
+                        bootstrapServers -> kafkaTracing.producer(TestUtils.producer(bootstrapServers)))
```
ocadaruma, line 103: Is that necessary to pass producerSupplier ?
```java
.producerDecorator(kafkaTracing::producer)
```
more natural?

brave/src/it/java/com/linecorp/decaton/processor/runtime/BraveTracingTest.java
```diff
@@ -0,0 +1,151 @@
...
+    public static class BraveTracePropagation implements ProcessingGuarantee {
+        private final Map<String, String> producedTraceIds = new ConcurrentHashMap<>();
+        private final Map<String, String> consumedTraceIds = new ConcurrentHashMap<>();
+        private final Tracing tracing;
+
+        public BraveTracePropagation(Tracing tracing) {this.tracing = tracing;}
+
+        @Override
+        public void onProduce(ProducedRecord record) {
+            producedTraceIds.put(record.task().getId(), tracing.currentTraceContext().get().traceIdString());
```
ocadaruma, line 58: Yeah, with brave-tracing we have to retrieve traceId like that.
But then I think it's bit confusing with `ProducedRecord#traceId`, it's only used with `TestTracingProvider`.

Also I feel it's bit inconsistent that `ProcessorTestSuite` always embed TestTracingProvider's trace id in record header, regardless of actual tracing provider.

So how about to do like following?
- Include `Headers` in `ProducedRecord` instead of `traceId`
- In processor module's `TracingTest`, retrieve traceId from header in `onProduce`
- Add `TestTracingProducer`, which generate and embed traceId in record header upon `send` and delegate to underlying producer
  * as like simple version of brave's TracingProducer
- In processor module's `TracingTest`, install `TracingProducer` in same way as `BraveTracingTest`

In this way, ProcessorTestSuite can be agnostic from concrete tracing provider at all, and we can use TestTracingProvider or any other tracing provider for tests in consistent way

ocadaruma (commented):
Thank you for the PR!

Yeah, tracing feature works in end-to-end from producer to processor, so integration test would be meaningful to check if it works correctly.

I added some feedbacks. Please take a look and give your opinions.

brave/src/it/java/com/linecorp/decaton/processor/runtime/BraveTracingTest.java
```diff
@@ -0,0 +1,151 @@
...
+                                        .retryTopic(retryTopic)
+                                        .backoff(Duration.ofMillis(10))
+                                        .producerSupplier(config -> kafkaTracing.producer(
+                                                producerSupplier.getProducer(config)))
+                                        .build())
+                .tracingProvider(new BraveTracingProvider(kafkaTracing))
+                // If we retry tasks, there's no guarantee about ordering nor serial processing
+                .excludeSemantics(
+                        GuaranteeType.PROCESS_ORDERING,
+                        GuaranteeType.SERIAL_PROCESSING)
```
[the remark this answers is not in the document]
m50d: Yes, per comment, since this test randomly retries half the tasks.

brave/src/it/java/com/linecorp/decaton/processor/runtime/BraveTracingTest.java
m50d: No, I was trying to diagnose why some traces were missing in a deployed server, but it turned out to be unrelated. Happy to remove.

brave/src/it/java/com/linecorp/decaton/processor/runtime/BraveTracingTest.java
m50d: I thought there might be future tests where we wanted to supply a completely new producer (e.g. a mock/stub, different configuration parameters) rather than just wrap the existing one.

brave/src/it/java/com/linecorp/decaton/processor/runtime/BraveTracingTest.java
m50d: Hmm. I worry that we're making test tracing work exactly the same way as brave tracing and will end up making assumptions that aren't valid for other tracing providers. But I don't have a better approach.

brave/src/it/java/com/linecorp/decaton/processor/runtime/BraveTracingTest.java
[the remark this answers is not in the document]
ocadaruma: Ah I misinterpreted. Yeah, that's totally right.

testing/src/main/java/com/linecorp/decaton/testing/processor/ProcessorTestSuite.java
```diff
@@ -87,6 +87,8 @@
     private final Set<ProcessingGuarantee> semantics;
     private final SubscriptionStatesListener statesListener;
     private final TracingProvider tracingProvider;
+    private final Function<String, Producer<String, DecatonTaskRequest>> producerSupplier;
+    private final Function<Integer, String> taskKeySupplier;
```
ocadaruma, line 91: Then we don't need `taskKeySupplier` for now?

ocadaruma (commented):
Added minor comment, but almost looks good.

m50d:
I think build failure is a known flaky test?

ocadaruma:
Oh, it's not known but shouldn't be related to this PR's change...
I created a ticket about that. https://github.com/line/decaton/issues/94

ocadaruma (approved):
LGTM, nice!

Issue #96, opened 2021-03-08
[Feature request] Timeout for `deferCompletion()`


techno:
We want a feature setting a timeout for `ProcessingContext.deferCompletion()` to avoid stuck `DecatonProcessor` in any case.

Decaton supports async task processing with using `deferCompletion()` on processor.
However, async processing usually makes a bug and it might miss to call `complete()`.
Finally, it causes critical issue of consuming topics.

If decaton has a timeout for `deferCompletion()`, we can avoid such outage in most cases.

kawamuray:
Thanks for your feedback :)

Yeah, after I saw several troubles that are caused by leaking a `DeferredCompletion`, it turns out that managing it properly and make sure to absolutely complete it is sometimes not an easy challenge and the consequence is really bad.

I'm +1 for adding kind of some support that lets users to configure forcing a `DeferredCompletion` to complete at some point even without an acknowledgement from the decaton processor.

Let me ask you few questions to consider the most suitable interface and implementation.

* When a `DeferredCompletion` completes forcefully, the corresponding record is considered as completed and the offset gets committed, so if it was not actually completed (failed) or the consumer instances crashes before it completes processing, it will lost. Is it fine for you?
* Is the timeout needs to be configured for each `deferCompletion()` call like `deferCompletion(Duration)` or, would it be sufficient to have single global timeout configured for each subscription?
* How long is it like, in practice you think you're going to set for the timeout?
* Do you think it will be useful if a user can select either to complete or retry the "expired" DeferredCompletion?

techno:
1. Yes, "pause" is more critical than "lost" for our workload. However, we want to track which task has been lost by timeout.
2. Global timeout for each subscription is enough for us.
3. It depends on topic, up to 1 min.
4. It seems very useful for retryable tasks.

kawamuray:
Thanks for answering. I got your feedback and will take it into account when designing this feature :)

kawamuray:
closing by https://github.com/line/decaton/pull/99/files

PR #99, opened 2021-03-18
Support deferred completion timeout as a workaround for "leak"

kawamuray:
For #96 

* Attempted to implement timeout (expiration) through offset state walking by OOOCC to avoid creating a new scheduled task per decaton task


processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ProcessingContextImpl.java
```diff
@@ -92,16 +91,23 @@ public String subscriptionId() {
     }
 
     @Override
-    public DeferredCompletion deferCompletion() {
+    public DeferredCompletion deferCompletion(Consumer<ProcessingContext<T>> timeoutCallback) {
```
ocadaruma, line 94: Hm, I wonder if this callback signature is confusing, because there are two `ProcessingContext`s inside the callback like below, so it's not clear which is appropriate. :
```java
void process(ProcessingContext context, Task task) {
    context.deferComplection(context2 -> {
        // there are context and context2
    });
}
```

Though in current implementation, they refers same `ProcessingContext`.
Just `Runnable` would be simple?

ocadaruma (commented):
Thanks for taking this!

I left some feedbacks, but overall looks good and seems works.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/Completion.java
```diff
@@ -0,0 +1,92 @@
...
+            if (!expire.dependency.tryExpire()) {
+                return false;
+            }
+            expire.dependency = null;
+        }
+
+        long now = System.currentTimeMillis();
+        if (expire.expireAt() <= now) {
+            expire.callback().run();
+            if (expire.expireAt() <= now) {
```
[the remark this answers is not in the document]
kawamuray: Yes, that's right.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ProcessingContextImpl.java
kawamuray: Hm, maybe yes. Let me think.

processor/src/main/java/com/linecorp/decaton/processor/ProcessingContext.java
```diff
@@ -60,9 +66,53 @@
...
+     * Note that you must return true from the callback even in this case to make sure that decaton waits until
+     * it completes retry-queuing asynchronously.
+     *
+     * if the callback returns false, decaton times out completion and forcefully completes it.
+     *
+     * @param callback callback which is called when the returned completion times out.
+     * @return a {@link Completion} which can be used to tell the completion of processing asynchronously.
      */
-    DeferredCompletion deferCompletion();
+    Completion deferCompletion(Supplier<Boolean> callback);
```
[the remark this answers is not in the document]
kawamuray: h-m, okay let me do that.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ProcessingContextImpl.java
```diff
@@ -98,16 +88,20 @@ public String subscriptionId() {
     }
 
     @Override
-    public DeferredCompletion deferCompletion() {
-        completionDeferred.set(true);
-        return completion;
+    public Completion deferCompletion(Supplier<Boolean> callback) {
+        if (deferredCompletion.get() == null) {
```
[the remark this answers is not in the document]
kawamuray: By creating a `CompletionImpl`  instance every time? I don't think so.

processor/src/main/java/com/linecorp/decaton/processor/ProcessingContext.java
```diff
@@ -60,9 +66,53 @@
...
+     */
+    default Completion deferCompletion() {
+        return deferCompletion(() -> false);
+    }
+
+    /**
+     * Tells the completion of this processing should be postponed and processor can accept next task.
+     * Once this method called within {@link DecatonProcessor#process} method, caller *MUST* call
+     * {@link DeferredCompletion#complete()} or {@link ProcessingContext#retry()} method in any cases.
+     * Otherwise consumption will stuck in short future and no new task will be given to the processor.
```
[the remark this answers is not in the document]
kawamuray: ok, let me add following sentence.

```
     * To avoid that with some risks, consider setting
     * {@link ProcessorProperties#CONFIG_DEFERRED_COMPLETE_TIMEOUT_MS}.
```

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/OffsetStateReaper.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public void maybeReapOffset(OffsetState state) {
+        if (state.completion().isComplete()) {
+            return;
+        }
+        long expireAt = state.expireAt();
+        long now = System.currentTimeMillis();
+        if (expireAt >= 0 && expireAt <= now) {
+            if (state.completion().tryExpire()) {
```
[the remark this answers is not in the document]
kawamuray: ogh, stupid mistake, thanks for a good catch!

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/OutOfOrderCommitControl.java
```diff
@@ -90,43 +73,17 @@ public synchronized DeferredCompletion reportFetchedOffset(long offset) {
...
         }
 
-        OffsetState state = new OffsetState(offset, false);
+        OffsetState state = new OffsetState(offset);
         states.addLast(state);
-        latest = state.offset;
+        latest = state.offset();
 
-        return () -> complete(state);
+        state.completion().future().whenComplete((unused, throwable) -> onComplete(offset)); // TODO okay in this order?
```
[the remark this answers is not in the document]
kawamuray: So this `whenComplete` call is the 1st one for this CF, and the CF will get several more `whenComplete` calls including ones supplied from user-supplied code. Since the these registered callbacks are called by LIFO order, I was wondering if logging at that last step of CF completion handling could cause misleading timestamp that is delayed by preceding callbacks.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/OffsetStateReaper.java
```diff
@@ -0,0 +1,53 @@
...
+                executor.execute(() -> {
+                    long nextExpireAt = now + completionTimeoutMs;
+                    state.setTimeout(nextExpireAt);
+                });
+            }
```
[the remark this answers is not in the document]
kawamuray: huh... it's supposed to be handled in `tryExpire` but looks like I removed that line at some point.

processor/src/main/java/com/linecorp/decaton/processor/runtime/Completion.java
```diff
@@ -0,0 +1,44 @@
...
+public interface Completion extends DeferredCompletion {
+    /**
+     * Returns whether this completion is complete.
+     * @return true if completed, false otherwise.
+     */
+    boolean isComplete();
+
+    /**
+     * Try to expire this completion. This method is intended for the use of decaton's internal implementation.
+     * @return true if successfully expired, false otherwise.
```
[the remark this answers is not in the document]
kawamuray: right, the doc reflects my intention so fixed the implementation.

processor/src/main/java/com/linecorp/decaton/processor/ProcessingContext.java
```diff
@@ -60,9 +67,57 @@
...
+     * {@link ProcessorProperties#CONFIG_DEFERRED_COMPLETE_TIMEOUT_MS}.
+     * You can do several things with the callback.
+     *
+     * If you know that the asynchronous processing is taking longer than expected but still running normally,
+     * thus want to tell decaton to extend timeout and let the processing complete, you can simply return true
+     * from the callback.
+     * {@code
+     * context.deferCompletion(() -> {
+     *     if (checkEverythingOk()) {
+     *         return true;
```
ocadaruma, line 97: Should fix example codes

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/CompletionImpl.java
```diff
@@ -0,0 +1,89 @@
...
+            return true;
+        }
+        if (dependency != null) {
+            if (!dependency.tryExpire()) {
+                return false;
+            }
+            dependency = null;
+        }
+        if (expireCallback.get() == TimeoutChoice.GIVE_UP) {
+            complete();
```
ocadaruma, line 68: Say we have processors like below:
```
- CompactionProcessor: deferCompletion(() -> GIVE_UP).completeWith(ctx.push())
  - UserProcessor1: deferCompletion(() -> EXTEND).completeWith(ctx.push())
    - UserPostProcessor2: deferCompletion(() -> GIVE_UP).completeWith(doSomethingOnAsyncFramework())
```

In current impl, since "leaf" processor's TimeoutChoice is GIVE_UP, expiration isn't extended when timeout happens, despite UserProcessor1 sets TimeoutChoice to EXTEND.

Is that intentional?

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/OffsetStateReaper.java
```diff
@@ -0,0 +1,53 @@
...
+    public void maybeReapOffset(OffsetState state) {
+        if (state.completion().isComplete()) {
+            return;
+        }
+        long expireAt = state.expireAt();
+        long now = System.currentTimeMillis();
+        if (expireAt >= 0 && expireAt <= now) {
+            long nextExpireAt = now + completionTimeoutMs;
+            executor.execute(() -> {
+                if (!state.completion().tryExpire()) {
```
ocadaruma, line 41: Since there's no place to set "root" completion (i.e. OffsetState#completion)'s expireCallback, this will cause NPE

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ProcessingContextImpl.java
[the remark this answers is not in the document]
ocadaruma: I wondered if we should check & set deferredCompletion atomically, since we could call `deferCompletion()` later from async-thread (i.e. ≠ processor thread), so it could cause a race.

But I realized that even in such case, deferCompletion should be called first in processor thread so current code will correctly do one-time initialization of the callback.

Okay, then fine

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/CompletionImpl.java
```diff
@@ -0,0 +1,89 @@
...
+import com.linecorp.decaton.processor.runtime.Completion;
+
+import lombok.Data;
+import lombok.experimental.Accessors;
+
+@Accessors(fluent = true)
+@Data
+public class CompletionImpl implements Completion {
+    private final CompletableFuture<Void> future;
+    private Supplier<TimeoutChoice> expireCallback;
```
ocadaruma, line 33: I think this should be volatile as well, since OffsetReaper thread may refer this while ProcessorThread calls setter.

ocadaruma (commented):
Added few more feedbacks.

Overall looks good.

processor/src/main/java/com/linecorp/decaton/processor/ProcessingContext.java
kawamuray: Indeed, good catch.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/OffsetStateReaper.java
kawamuray: Good point.. Also realized while writing unit tests. Fixed in the following commit.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/CompletionImpl.java
kawamuray: No... In that case, after 2 is expired 1's callback should be called, hence it should extend timeout. Let me fix that, super nice catch.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/CompletionImpl.java
kawamuray: True, gonna fix that.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/CompletionImpl.java
```diff
@@ -0,0 +1,95 @@
...
+    @Override
+    public boolean tryExpire() {
+        if (isComplete()) {
+            return true;
+        }
+        if (dependency != null) {
+            if (!dependency.tryExpire()) {
+                return false;
+            }
+            dependency = null;
```
ocadaruma, line 67: (Just a confirmation) Say we have processors like below:

```
- CompactionProcessor: deferCompletion(() -> GIVE_UP).completeWith(ctx.push())
  - UserProcessor1: deferCompletion(() -> EXTEND).completeWith(ctx.push())
    - UserPostProcessor2: deferCompletion(() -> GIVE_UP).completeWith(doSomethingOnAsyncFramework())
```

When completion timed out once, since `UserProcessor2`'s choice is GIVE_UP, `UserProcessor1`'s dependency will be substituted with null.

Since `UserProcessor1`'s choice is EXTEND, the timeout will be extended.

When the timeout happens 2nd time, since `UserProcessor1`'s dependency is already null, `UserProcessor2`'s callback will not be called.

But it is expected and okay, because `UserProcessor2` chose GIVE_UP on first timeout, it means that we don't have to take `UserProcessor2`'s choice into account of expiration handling.

Is my understanding correct?

ocadaruma (commented):
Added nits comment and question, but overall looks good.

Besides, isn't useful to expose a metric to observe timed-out tasks?

processor/src/main/java/com/linecorp/decaton/processor/runtime/Completion.java
```diff
@@ -0,0 +1,58 @@
...
+package com.linecorp.decaton.processor.runtime;
+
+import java.util.concurrent.CompletionStage;
+
+import com.linecorp.decaton.processor.DeferredCompletion;
+
+/**
+ * An interface for completion status management of tasks.
+ */
+public interface Completion extends DeferredCompletion {
```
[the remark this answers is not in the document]
kawamuray: Yeah, I thought about the same thing but decided to leave it because `DeferredCompletion` must be one of the interfaces that are referred often by user code, so even though we're going to bump major version, removing it will cause a lot of user applications to stop compiling.

I'm not sure what would be the best way here. I think keeping it is harmless in terms of maintainability, but if prefer to avoid leaving such garbage which has no goodenough reason to keep, we may mark it `@Deprecated` in this release and remove at some point in future?

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/CompletionImpl.java
kawamuray: That's correct.

processor/src/main/java/com/linecorp/decaton/processor/runtime/Completion.java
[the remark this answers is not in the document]
ocadaruma: > DeferredCompletion must be one of the interfaces that are referred often by user code

Hm maybe.

Then sounds fine to keep the interface for now.

processor/src/main/java/com/linecorp/decaton/processor/DeferredCompletion.java
```diff
@@ -35,4 +36,16 @@
...
         return future.whenComplete((r, e) -> complete());
     }
+
+    /**
+     * Associate completion of this completion to the given {@link Completion}.
+     * By calling this method for the completion X with the argument of A,
+     * * when A completes, X completes
+     * * when X times out, A times out
+     *
+     * Always prefer to use this method instead of {@link #completeWith(CompletableFuture)} when possible.
```
ocadaruma, line 46: > Always prefer to use this method instead of {@link #completeWith(CompletableFuture)} when possible.

I think users (also me :) ) may wonder why is that.
Could you elaborate why in the comment, or just delete the sentence?

ocadaruma (commented):
Left one comment but I think we almost reaching the goal..!

Also Let me ask two points:

- WDYT about this?
  * > Besides, isn't useful to expose a metric to observe timed-out tasks?
- It's definitely worth to describe this feature in Decaton doc. Will it be done in another PR?

kawamuray:
> Besides, isn't useful to expose a metric to observe timed-out tasks?

oops I missed that, good idea, let me do that.

> It's definitely worth to describe this feature in Decaton doc. Will it be done in another PR?

Yeah maybe better to be done a bit later.
* we would want to confirm applicability of this feature in real-world application first
* the doc would not just mention about this feature but should probably describe how to use it, maybe like "async-processing in decaton" overall?

ocadaruma (commented):
Test is failing

ocadaruma (approved):
LGTM. great work!

PR #102, opened 2021-04-23
Stop setting record timestamp from meta fixes #101

### LINKED ISSUE #101: ProducerRecord for retry tasks could be generated with timestamp=0
- When we consume non-DecatonTaskRequest topic with `TaskExtractor`, `TaskMetadata#timestampMillis` could be set to 0 (i.e. long's default value) if we don't set it explicitly.
- Retry tasks inherits original task's metadata https://github.com/line/decaton/blob/v1.1.0/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/DecatonTaskRetryQueueingProcessor.java#L57
- DecatonTaskProducer sets `ProducerRecord#timestamp` from `TaskMetadata`
  * https://github.com/line/decaton/blob/v1.1.0/client/src/main/java/com/linecorp/decaton/client/internal/DecatonTaskProducer.java#L69
- Thus, retry tasks will be produced with `ProducerRecord#timestamp=0`

This could cause the messages to be deleted on the broker side unexpectedly earlier, since retention-based log deletion is happened based on record's timestamp. (https://kafka.apache.org/documentation/#impl_deletes)

ocadaruma:
## Summary
- As described in #101, when we don't set `TaskMetadata#timestampMillis` explicitly on `TaskExtractor`, retry record's timestamp could be set to 0, which cause records to be deleted on broker side by retention immediately
- As retention-based log deletion takes record's timestamp into account, we should supply it independently of task's metadata timestamp (i.e. which is expected to be not related to Kafka's internal)

## Changes
- Stop setting record's timestamp from task metadata in DecatonTaskProducer.
  * We leave record's timestamp as null, so that KafkaProducer fills `System.currentTimeMillis`
  * NOTE: This could be considered as "backward-incompatible change", so we should bump major version when releasing this fix
- Add an integration test to reproduce the phenomenon (i.e. retry tasks are produces with timestamp=0 unintentionally, then deleted by retention immediately)

### DIFF
--- a/client/src/main/java/com/linecorp/decaton/client/internal/DecatonTaskProducer.java
+++ b/client/src/main/java/com/linecorp/decaton/client/internal/DecatonTaskProducer.java
@@ -18,7 +18,6 @@ package com.linecorp.decaton.client.internal;
 
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Properties;
 import java.util.concurrent.CompletableFuture;
 
@@ -30,7 +29,6 @@ import com.linecorp.decaton.client.DecatonClient;
 import com.linecorp.decaton.client.KafkaProducerSupplier;
 import com.linecorp.decaton.client.PutTaskResult;
 import com.linecorp.decaton.protocol.Decaton.DecatonTaskRequest;
-import com.linecorp.decaton.protocol.Decaton.TaskMetadataProto;
 
 /**
  * A raw interface to put a built {@link DecatonTaskRequest} directly.
@@ -64,9 +62,7 @@ public class DecatonTaskProducer implements AutoCloseable {
     }
 
     public CompletableFuture<PutTaskResult> sendRequest(String key, DecatonTaskRequest request) {
-        TaskMetadataProto taskMeta = Objects.requireNonNull(request.getMetadata(), "request.metadata");
-        ProducerRecord<String, DecatonTaskRequest> record =
-                new ProducerRecord<>(topic, null, null, key, request);
+        ProducerRecord<String, DecatonTaskRequest> record = new ProducerRecord<>(topic, key, request);
 
         CompletableFuture<PutTaskResult> result = new CompletableFuture<>();
         producer.send(record, (metadata, exception) -> {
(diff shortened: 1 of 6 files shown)


client/src/main/java/com/linecorp/decaton/client/internal/DecatonTaskProducer.java
```diff
@@ -66,7 +66,7 @@ public DecatonTaskProducer(String topic, Properties producerConfig,
     public CompletableFuture<PutTaskResult> sendRequest(String key, DecatonTaskRequest request) {
         TaskMetadataProto taskMeta = Objects.requireNonNull(request.getMetadata(), "request.metadata");
         ProducerRecord<String, DecatonTaskRequest> record =
-                new ProducerRecord<>(topic, null, taskMeta.getTimestampMillis(), key, request);
+                new ProducerRecord<>(topic, null, null, key, request);
```
kawamuray, line 69: Just use another overload? `public ProducerRecord(String topic, K key, V value)`

processor/src/it/java/com/linecorp/decaton/processor/RetryQueueingTest.java
```diff
@@ -99,4 +134,46 @@ public void testRetryQueuing() throws Exception {
...
+                .propertySupplier(StaticPropertySupplier.of(
+                        // Configure max pending records to 1 to pause fetches easily
+                        Property.ofStatic(ProcessorProperties.CONFIG_MAX_PENDING_RECORDS, 1),
+                        Property.ofStatic(ProcessorProperties.CONFIG_PARTITION_CONCURRENCY, 1)
+                ))
+                .configureProcessorsBuilder(builder -> builder.thenProcess((ctx, task) -> {
+                    if (ctx.metadata().retryCount() == 0) {
+                        ctx.retry();
+                    } else {
+                        Thread.sleep(1000L);
```
kawamuray, line 164: Assuming the purpose of this sleep is to let broker enough time to run retention check of log segments, is it fine to apply it just for the first record so the overall test duration won't get long?

kawamuray: btw this needs to be sufficient larger than https://github.com/line/decaton/pull/102/files#diff-50a812e0953575d473c752613b37560bdd78696bae7e051019e48e21d278029eR60 this right? How about definining a constant and use it so the relationship will be clearer?

kawamuray (changes requested):
Thanks for working on this :) Just few minor points.

processor/src/it/java/com/linecorp/decaton/processor/RetryQueueingTest.java
ocadaruma: > is it fine to apply it just for the first record

Sounds enough. Let me fix

kawamuray (approved):
LGTM 👍

PR #104, opened 2021-05-11
Reduce some object allocations that are identified as expensive by allocation profiling

kawamuray:
I found some code in decaton core causes massive object allocations (in proportional to consumed records) that causes high GC pressure.

The most significant ones are String allocation in `LoggingContext` (`TaskMetadata.toString()`) but `TaskRequest#id` too. (which is used only if trace level logging is enabled)

In this PR I saved those expensive object allocations by making it lazy or by adding a new flag to opt-out `LoggingContext` when its not necessary. 

* Adds new config `decaton.logging.mdc.enabled`

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/LoggingContext.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/LoggingContext.java
@@ -38,22 +38,29 @@ public class LoggingContext implements AutoCloseable {
     public static final String SUBSCRIPTION_ID_KEY = "dt_subscription_id";
     public static final String TASK_KEY = "dt_task_key";
 
-    public LoggingContext(String subscriptionId, TaskRequest request, TaskMetadata metadata) {
-        MDC.put(METADATA_KEY, metadata.toString());
-        MDC.put(TASK_KEY, String.valueOf(request.key()));
-        MDC.put(SUBSCRIPTION_ID_KEY, subscriptionId);
-        MDC.put(OFFSET_KEY, String.valueOf(request.recordOffset()));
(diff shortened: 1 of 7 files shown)


processor/src/main/java/com/linecorp/decaton/processor/LoggingContext.java
```diff
@@ -38,22 +38,29 @@
...
-    public LoggingContext(String subscriptionId, TaskRequest request, TaskMetadata metadata) {
-        MDC.put(METADATA_KEY, metadata.toString());
-        MDC.put(TASK_KEY, String.valueOf(request.key()));
-        MDC.put(SUBSCRIPTION_ID_KEY, subscriptionId);
-        MDC.put(OFFSET_KEY, String.valueOf(request.recordOffset()));
-        MDC.put(TOPIC_KEY, request.topicPartition().topic());
-        MDC.put(PARTITION_KEY, String.valueOf(request.topicPartition().partition()));
+    private final boolean enabled;
+
+    public LoggingContext(boolean enabled, String subscriptionId, TaskRequest request, TaskMetadata metadata) {
```
ocadaruma, line 43: I think it's better to keep current constructor to avoid unnecessary breaking change since this class is considered as public API.

ocadaruma (commented):
Only left one comment, but almost looks good.

processor/src/main/java/com/linecorp/decaton/processor/LoggingContext.java
kawamuray: Good point. Let me follow that.

ocadaruma (approved):
LGTM!

PR #105, opened 2021-05-12
Make max.poll.records customizable through consumer properties

kawamuray:
In recent performance tuning experience, I found that the current default max.poll.records cap could be an issue.
As Decaton expects I/O intensive workload, which typically takes long to process a task, often the cost of work in Consumer#poll is negligible so we added hard limit for max.poll.records to size OOOCC easily.

However, in processor that does large batching of tasks, the processing time of a task could be few microseconds and in case `Consumer#poll` becomes an bottleneck. For such case, increasing max.poll.records contributes to better performance by eliminating some works repeated on every call of `Consumer#poll.`

This PR makes it to regard user-supplied max.poll.records in consumer config and discard hard limit, by using user-supplied max.poll.records to compute the size of OOOCC, when it's supplied.

This could be potentially a breaking change, but I would say no because the expected beaviour change has no critical consequence.

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/SubscriptionBuilder.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/SubscriptionBuilder.java
@@ -143,6 +143,14 @@ public class SubscriptionBuilder {
         return this;
     }
 
+    private int consumerMaxPollRecords() {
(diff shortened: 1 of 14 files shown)


ocadaruma (approved):
Good point.

LGTM.

> This could be potentially a breaking change, but I would say no because the expected beaviour change has no critical consequence.

Agree. I don't think this is a breaking change since I guess no one is relying on current behavior that "if max.poll.records > 100 it's ignored and 100 is set forcibly" .

PR #106, opened 2021-05-19
Upgrade micrometer to 1.6.6

mauhiz:
**Motivation:** Micrometer 1.7 is out and we shouldn't stay 2 minor versions behind

I refrained from jumping right to 1.7 because
- it's still quite recent
- it introduces a breaking change by requiring prometheus simpleclient 0.10.0, which has a small change in metrics name convention. https://github.com/prometheus/client_java/releases/tag/parent-0.10.0


kawamuray:
LGTM, thanks!

PR #108, opened 2021-06-07
default linger.ms 0 doesn't make sense typically

ocadaruma:
## Motivation
- Retry producer will be instantiated with `linger.ms = 0` unless we explicitly set, since it's kafka-clients's default.
- But typically it doesn't make much sense because we don't need such super short delivery latency for retry producer.

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/SubscriptionBuilder.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/SubscriptionBuilder.java
@@ -48,15 +48,15 @@ import lombok.experimental.Accessors;
 @Setter
 @Accessors(fluent = true)
 public class SubscriptionBuilder {
-    private static final Map<String, String> presetProducerConfig;
+    private static final Map<String, String> presetRetryProducerConfig;
 
     static {
-        presetProducerConfig = new HashMap<>();
+        presetRetryProducerConfig = new HashMap<>();
         // In Decaton processor which handles massive traffic, retry tasks could cause
         // production burst when processes start to fail. (e.g. due to downstream service down)
         // Since default producer's linger.ms is 0, this could harm Kafka cluster despite we
         // don't much care about retry task's delivery latency typically. So we set reasonable default here.
-        presetProducerConfig.put(ProducerConfig.LINGER_MS_CONFIG, "100");
+        presetRetryProducerConfig.put(ProducerConfig.LINGER_MS_CONFIG, "100");
     }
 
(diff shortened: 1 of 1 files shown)


processor/src/main/java/com/linecorp/decaton/processor/runtime/SubscriptionBuilder.java
```diff
@@ -166,8 +166,15 @@ public ProcessorSubscription build() {
...
         DecatonProcessorSupplier<byte[]> retryProcessorSupplier = null;
         if (retryConfig != null) {
-            Properties producerConfig = Optional.ofNullable(retryConfig.producerConfig())
-                                                .orElseGet(producerConfigSupplier(consumerConfig));
+            Properties producerConfig = new Properties();
+            // In Decaton processor which handles massive traffic, retry tasks could cause
+            // production burst when processes start to fail. (e.g. due to downstream service down)
+            // Since default producer's linger.ms is 0, this could harm Kafka cluster despite we
+            // don't much care about retry task's delivery latency typically. So we set reasonable default here.
+            producerConfig.setProperty(ProducerConfig.LINGER_MS_CONFIG, "100");
```
kawamuray, line 174: Can you pull this up to somewhere like https://github.com/line/decaton/blob/master/client/src/main/java/com/linecorp/decaton/client/internal/DecatonTaskProducer.java#L42 for ability to easy spot where such defaults are defined.

kawamuray (changes requested):

processor/src/main/java/com/linecorp/decaton/processor/runtime/SubscriptionBuilder.java
ocadaruma: I thought about that, but I think making such high linger as the default is only acceptable for retry producers.

But yeah, defining in-line could make spotting the default difficult.
Let me find better way to define defaults for retry producer.

processor/src/main/java/com/linecorp/decaton/processor/runtime/SubscriptionBuilder.java
```diff
@@ -46,6 +48,17 @@
 @Setter
 @Accessors(fluent = true)
 public class SubscriptionBuilder {
+    private static final Map<String, String> presetProducerConfig;
```
kawamuray, line 51: `preset*Retry*ProducerConfig`?

kawamuray (changes requested):

kawamuray (approved):
LGTM, thanks!

PR #110, opened 2021-06-08
Support MDC enable/disable flag in BM for allocation profiling

kawamuray:


ocadaruma (approved):
Nice. As enabling MDC is known to increase allocation certainly

PR #113, opened 2021-06-08
Correct package of Completion class as it's required to construct a processor

kawamuray:
Follow-up for #99


ocadaruma (approved):
Thanks. LGTM

PR #114, opened 2021-06-09
Fix typo

ocadaruma:


kawamuray:
lol, nice catch, thanks

kawamuray:
This is confusing... semver started from 0.1.0 but we haven't created a release, so it looks like 1.0.0 is the first release with semver.... well whatever, let's merge then

PR #116, opened 2021-06-21
Properly shutdown subscription even when its closed without started

### LINKED ISSUE #115: Waiting indefinitely if closing a unstarted ProcessorSubscription
`close()` will wait indefinitely if closing a `ProcessorSubscription` without calling `start()`.
This symptom occurs from version `1.1.0`.

If I don't need to start it, I can implement it so that it doesn't create a `ProcessorSubscription`, but I thought this symptom was strange.

Here is the simple reproducer
```java
var properties = new Properties();
properties.setProperty(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, "<bootstrap-servers>");
properties.setProperty(ConsumerConfig.CLIENT_ID_CONFIG, "test");
properties.setProperty(ConsumerConfig.GROUP_ID_CONFIG, "test");
var subscription = SubscriptionBuilder.newBuilder("my-decaton-processor").processorsBuilder(
                ProcessorsBuilder.consuming("my-decaton-topic", new ProtocolBuffersDeserializer<>(SomeTask.parser())))
                                                                .consumerConfig(properties).build();
subscription.close(); // hang-up...
```

kawamuray:
Fixes #115

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
@@ -22,7 +22,6 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -57,7 +56,6 @@ import lombok.extern.slf4j.Slf4j;
 @Slf4j
 public class ProcessorSubscription extends Thread implements AsyncShutdownable {
     private final SubscriptionScope scope;
-    private final AtomicBoolean terminated;
     private final BlacklistedKeysFilter blacklistedKeysFilter;
     final PartitionContexts contexts;
     private final Processors<?> processors;
@@ -67,7 +65,10 @@ public class ProcessorSubscription extends Thread implements AsyncShutdownable {
     private final CommitManager commitManager;
     private final AssignmentManager assignManager;
     private final ConsumeManager consumeManager;
-    private final CompletableFuture<Void> shutdownFuture = new CompletableFuture<>();
+    private final CompletableFuture<Void> loopTerminateFuture;
+    private final CompletableFuture<Void> shutdownFuture;
+    private volatile boolean started;
+    private volatile boolean terminated;
 
     class Handler implements ConsumerHandler {
         @Override
@@ -136,7 +137,6 @@ public class ProcessorSubscription extends Thread implements AsyncShutdownable {
         this.processors = processors;
         this.stateListener = stateListener;
         this.contexts = contexts;
-        terminated = new AtomicBoolean();
         metrics = Metrics.withTags("subscription", scope.subscriptionId()).new SubscriptionMetrics();
 
         Consumer<String, byte[]> consumer = consumerSupplier.get();
@@ -146,6 +146,9 @@ public class ProcessorSubscription extends Thread implements AsyncShutdownable {
         assignManager = new AssignmentManager(contexts);
         blacklistedKeysFilter = new BlacklistedKeysFilter(props);
         rebalanceTimeoutMillis = props.get(ProcessorProperties.CONFIG_GROUP_REBALANCE_TIMEOUT_MS);
+
+        loopTerminateFuture = new CompletableFuture<>();
+        shutdownFuture = loopTerminateFuture.whenComplete((unused, throwable) -> cleanUp());
 
         setName(String.format("DecatonSubscriptionThread-%s", scope));
     }
@@ -212,11 +215,21 @@ public class ProcessorSubscription extends Thread implements AsyncShutdownable {
 
     @Override
     public void run() {
-        updateState(SubscriptionStateListener.State.INITIALIZING);
-        consumeManager.init(subscribeTopics());
-
+        started = true;
         try {
-            while (!terminated.get()) {
+            if (!terminated) {
+                updateState(SubscriptionStateListener.State.INITIALIZING);
+                consumeManager.init(subscribeTopics());
+                consumeLoop();
+            }
+        } finally {
+            loopTerminateFuture.complete(null);
+        }
+    }
+
+    private void consumeLoop() {
+        try {
+            while (!terminated) {
                 consumeManager.poll();
 
                 Timer timer = Utils.timer();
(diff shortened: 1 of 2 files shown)


processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -246,24 +261,26 @@ public void run() {
             } catch (RuntimeException e) {
                 log.error("Offset commit failed before closing consumer", e);
             }
-            contexts.close();
-
             processors.destroySingletonScope(scope.subscriptionId());
```
ocadaruma, line 264: If we keep this line, `processors.destroySingletonScope(scope.subscriptionId());` will be called twice upon close. Is that safe?

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -246,24 +261,26 @@ public void run() {
...
-
-            consumeManager.close();
-            log.info("ProcessorSubscription {} terminated in {} ms", scope,
-                     timer.elapsedMillis());
-
-            updateState(SubscriptionStateListener.State.TERMINATED);
-            metrics.close();
-            shutdownFuture.complete(null);
+            log.info("ProcessorSubscription {} terminated in {} ms", scope, timer.elapsedMillis());
```
ocadaruma, line 265: There are several procedures which might take certain time after this, until complete termination at this line. https://github.com/line/decaton/pull/116/files#diff-37fa89e1da134b048d335ebf4753a63374e56c59dd65882fc1e84ac1aad44cacR283

Shouldn't we log when terminated completely?

ocadaruma (commented):
Thanks for fixing. Left few comments.

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
kawamuray: Good catch... I was intend to leave this and do not include it in `cleanUp`

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
kawamuray: Yeah, thought about something similar, but concluded to leave this here because of:

* This message intends to tell the time taken to terminate subscription "thread"
* likewise, I think it's strange to see "terminated ..." message if one called `.close()` on subscription w/o starting it
* because the message here depends on the timer, we have to to bring it somehow to `cleanUp` which is unnecessarily complex.

kawamuray: In contrast, I put state change notification for `TERMINATED` in real-end because I thought it could be used for some purpose to trigger some processsing after exactly everything of this subscription to have destroyed, so I thought it could be bit danger to call it at the end of thread termination.

ocadaruma: > This message intends to tell the time taken to terminate subscription "thread"

Hmm, even "thread"'s termination not completed yet at this line I think? ("thread is terminated" == "run() returns" right?)

So better to log after `run` returns ?

> because the message here depends on the timer, we have to to bring it somehow

I see, that's right.
Hm, then let's just get rid of timer and add another log at the start of termination procedure?
We can see the time to the shutdown by log timestamp

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -246,24 +261,25 @@ public void run() {
             } catch (RuntimeException e) {
                 log.error("Offset commit failed before closing consumer", e);
             }
-            contexts.close();
-
             processors.destroySingletonScope(scope.subscriptionId());
```
ocadaruma, line 264: H-m, better to move to `cleanUp` ?
I think there's a race like below:

Say user is enabled DC timeout and retry feature.

1. `processors.destroySingletonScope(scope.subscriptionId());`
    * retry processors accepts no more tasks from now on
2. timeout callback (executed on OffsetStateReaper thread) is taking long time (for user-supplied logic)
3. after a while, timeout callback decides the task to retry, then executed `ctx.retry()`
4. this spits garbage ERROR log because producer is already closed

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -212,11 +215,21 @@ private void updateState(SubscriptionStateListener.State newState) {
 
     @Override
     public void run() {
-        updateState(SubscriptionStateListener.State.INITIALIZING);
-        consumeManager.init(subscribeTopics());
+        started = true;
+        try {
+            if (!terminated) {
```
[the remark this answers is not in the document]
kawamuray: There's potential race such that:

* A: the subscription thread to execute `run()`
* B: the thread calling `.close()` (or it's siblings)

* B: `terminated = true`
* A: `started = true`
* A:  calls `consumer.subscribe()`
* B: do delegate completion of `loopTerminateFuture` to A

in such case, without this check the process still can be cleaned up, but because of A can be already ware of termination flag, I thought running `consumer.subscribe()` which then involves partition assignments => processors creation is not negligibly time-consuming work to let it proceed.
So I thought checking terminated at the very top of the `run()` would be worth.

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
kawamuray: > Hmm, even "thread"'s termination not completed yet at this line I think? ("thread is terminated" == "run() returns" right?)

That's true, but the remaining procedure are negligible cheap and are supposed to finish very quickly isn't it?
If the purpose of this logging is to tell the exact timing of `ProcessorSubscription`'s destroy, I would agree with you, but I think the purpose of this log is to tell how long did it take to complete the cleanup sequence where the main part comes from processor's destruction.

>  Hm, then let's just get rid of timer and add another log at the start of termination procedure?

That works, but less convenient than as-is because users have to find two lines and compute duration by themselves.
I always prefer kind logging as possible as we can, so this style still fulfils the purpose IMO. 

If you think this log message gonna be confusing, shall we change it to be more precises like "ProcessorSubscription thread {} cleaned up in {} ms" ?

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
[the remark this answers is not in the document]
ocadaruma: Thanks for explanation.
Makes sense.

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
ocadaruma: We found that this race also could happen for user-provided processors, because we can't ensure no more retry tasks are queued after destroying retry processor.
e.g:
1. processors.destroySingletonScope(scope.subscriptionId());
    * retry processors accepts no more tasks from now on
2. some tasks are still executed asynchronously on some thread
3. eventually `ctx.retry()` is called. but retry processor is already closed

So, current PR's execution order (destroySingletonScope => contexts.close()) doesn't change semantics much at least.

So current impl looks fine.

ocadaruma (approved):

Issue #98, opened 2021-03-18
Support cooperative rebalancing protocol


ocadaruma:
- Kafka 2.4.0 introduces `cooperative rebalancing protocol`, which mitigates the stop-the-world effect upon rebalancing
  * https://cwiki.apache.org/confluence/display/KAFKA/KIP-429%3A+Kafka+Consumer+Incremental+Rebalance+Protocol
- Users can opt-in cooperative rebalancing by specifying `CooperativeStickyAssignor` in `partition.assignment.strategy` config or implementing `ConsumerPartitionAssignor` which returns `RebalanceProtocol.COOPERATIVE` from `supportedProtocols`
- If cooperative rebalancing is enabled, Kafka consumer will invoke `ConsumerRebalanceListener` as like below:
  * `onPartitionsRevoked` : called only when there's at least one partition that revoked from current assignment
  * `onPartitionsAssigned`: called with **newly** assigned partitions
  * refs: https://cwiki.apache.org/confluence/display/KAFKA/KIP-429%3A+Kafka+Consumer+Incremental+Rebalance+Protocol#KIP429:KafkaConsumerIncrementalRebalanceProtocol-ConsumerRebalanceListenerandConsumerPartitionAssignorSemantics
- `onPartitionsAssigned`'s behavior change makes Decaton to not working, because `AssignmentManager` assumes partitions which previously assigned and kept assigned after rebalance as `revoked` unintentionally.
  * https://github.com/line/decaton/blob/master/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/AssignmentManager.java#L109

ocadaruma:
Cooperative assignor will be used by default from 3.3.0 https://cwiki.apache.org/confluence/display/KAFKA/Release+Plan+3.3.0

PR #120, opened 2021-07-21
Adopt poll(Duration) fixes #118

### LINKED ISSUE #118: Use poll(Duration)
## Summary
- Currently, Decaton uses `KafkaConsumer#poll(long)` which could block indefinitely and already marked as deprecated
  * https://github.com/line/decaton/blob/v3.0.1/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ConsumeManager.java#L165
- We should migrate to `poll(Duration)` overload

ocadaruma:
## Summary
- Currently, Decaton uses `KafkaConsumer#poll(long)` which is deprecated and could be removed in future Kafka release

### DIFF
--- a/processor/src/it/java/com/linecorp/decaton/processor/SubscriptionStateTest.java
+++ b/processor/src/it/java/com/linecorp/decaton/processor/SubscriptionStateTest.java
@@ -26,13 +26,35 @@ import java.util.Deque;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


ocadaruma:
Seems like there's behavior change regarding rebalance-listener invocation.
Needs consideration how to address that.

ocadaruma:
I'll address this in another PR to fix #98.
Closing this PR for now

PR #122, opened 2021-08-12
Use toBuilder = true on config classes

m50d:
Small change to make it easer to reuse configs while changing some properties. (Specifically, this is useful for the case where we want to populate a `RetryConfig` automatically from Spring properties, but then set some fields programmatically)


kawamuray:
LGTM, thanks!

PR #124, opened 2021-09-13
Leave completion object in ProcessingContext field to avoid leak

kawamuray:


ocadaruma (approved):
Thanks for quick fixing.

Roughly checked locally that no completion-leak even we call ctx.deferCompletion() multiple times.

ocadaruma:
Let me send follow-up PR to add integration tests about this.

PR #125, opened 2021-09-14
Add tests for deferred completion

ocadaruma:
## Motivation

- In https://github.com/line/decaton/pull/99, we made some spec change about `ProcessingContext#deferCompletion()` that:
  * It requires to re-use `Completion` instance returned from `deferCompletion()` if we want to refer it later, otherwise completion will be leaked
  * However we realized that this change could cause a lot of migration cost for Decaton <= 2.x users
      - Decaton doc's example code was also outdated
  * So we concluded to make keep existing deferCompletion-usage to work by adding fix (https://github.com/line/decaton/pull/124)
  * => We should add integration test cases to ensure it works
- Besides, I fixed `ProcessorTestSuite` to await retried-produced tasks to be committed


kawamuray (changes requested):

kawamuray (approved):
LGTM, thanks 👍

PR #132, opened 2021-11-22
Allow to bind KafkaClientMetrics for each decaton consumer

mauhiz:
## Motivation

- `micrometer` 1.4+ has deprecated the JMX-based based metrics gathering ( `KafkaConsumerMetrics` ) in favour of converting `Consumer` exposed `Metrics`. See [this PR](https://github.com/micrometer-metrics/micrometer/pull/1835).
- decaton metrics don't cover some of the generic consumer metrics like records lag

## Changes

Always bind `KafkaClientMetrics`, tagged with the right subscription id.
This should not be a breaking change since there is no reasonable way to have already bound `KafkaClientMetrics` for these consumers elsewhere.
Note : not doing this for `KafkaProducerSupplier` which is overridable / customizable.

(I'd like advice on how to deal with the test failure - I don't understand why we have to manage micrometer registry cleanup manually)


kawamuray:
Thanks for PR!

I think one of the downsides of enabling `KafkaClientMetrics` by default is it would explode number of metrics in registry provided by decaton?
Decaton provides its own registry to let users easy control what to do with it, like exposing directly, or combining it with their own registry to mix the resulting metrics set.
While I think most of users will need kafka client's metrics in addition to decaton's metrics, I thought most of them using `metrics.reporters` to bypass decaton layers and inject client directly their own reporter, which sounds more suitable way than decaton to take care of it, but micormeter apparently doesn't provide out-of-the-box support for using `metrics.reporters`?
In case, I think supporting `KafkaClientMetrics` in decaton is good, but would it be better to expose `ConsumerSupplier` interface instead? or at least a flag to enable (disable) `KafkaClientMetrics` binding is needed IMO.


> (I'd like advice on how to deal with the test failure - I don't understand why we have to manage micrometer registry cleanup manually)

That's because each decaton subscription creates metrics with its own label sets, so unless we clean these metrics up at the end of subscription's lifetime, there will be a garbage created which will keep accumulating. It wouldn't become a significant problem in  most of decaton use-cases where decaton subscription dies along with application instance, but still should be cleaned up as possible as we can as a manner.

mauhiz:
> is it would explode number of metrics in registry provided by decaton?

yes, assuming people don't already monitor their Kafka client via JMX adapted (namely, micrometer's deprecated `KafkaConsumerMetrics`  ).

> micrometer apparently doesn't provide out-of-the-box support for using metrics.reporters

If I understand what you mean correctly, isn't that exactly the `KafkaClientMetrics` this PR is introducing?

> expose `ConsumerSupplier`

My first thought, but it was not convenient since it doesn't have the `subscriptionId` (and not sure it should). 

> flag to enable/disable `KafkaClientMetrics`

Sounds good. Will add this flag.

> clean these metrics up at the end of subscription's lifetime

I see, if we go the way I suggest in this PR I will fix to cleanup the metrics
What do other maintainers think?

mauhiz:
I couldn't find a way to cleanup `KafkaClientMetrics` on subscription close. So, making it a more free option by exposing a builder for it, and leaving the binding up to the user.

kawamuray:
> If I understand what you mean correctly, isn't that exactly the KafkaClientMetrics this PR is introducing?

No, they favored the way using `#metrics()` method of `Consumer` to obtain metrics collection and wrap it over. What I said is to use `metric.reporters` property, which you specify the class name that implements `MetricsReporter` interface, so you can give it through properties you give to `SubscriptionBuilder#consumerConfig()`.

kawamuray:
> I couldn't find a way to cleanup KafkaClientMetrics on subscription close. 

Indeed with the current version, but looks like it has added in latest?

https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/kafka/KafkaMetrics.java#L295

mauhiz:
> `metric.reporters` property

Okay, I didn't know there was such a thing. I don't know why would they use or not use it :D

> added in latest

a-ha, nice. Let me try to integrate it.

mauhiz:
PTAL. 
Note that these new metrics are prefixed with `decaton.` but that is fine, since they have an extra dimension (subscriptionId) they are not mergeable with generic consumer metrics

mauhiz:
wow, many tests started timing out... why..

kawamuray:
Just guessing, by increased number of metrics ...?  We might need profiling to be sure.

mauhiz:
It turns out that calling `properties` in `TestUtils` would override the properties defined right before in `builderConfigurer.accept(builder)`.

I added commits to
- avoid overwriting `properties` in SubscriptionBuilder - I can't see any good reason to do it
- add timeouts of 20s in some places of the tests that were blocking forever, for easier debugging

processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
```diff
@@ -99,6 +102,23 @@ public void close() {
     }
 
     public class SubscriptionMetrics extends AbstractMetrics {
+        volatile KafkaClientMetrics kafkaClientMetrics;
+
+        public void bindClientMetrics(Consumer<String, byte[]> consumer) {
+            if (kafkaClientMetrics != null) {
+                closeClientMetrics();
```
kawamuray, line 109: In which case would this be effective?

processor/src/main/java/com/linecorp/decaton/processor/runtime/SubscriptionBuilder.java
```diff
@@ -59,8 +59,8 @@
         presetRetryProducerConfig.put(ProducerConfig.LINGER_MS_CONFIG, "100");
     }
 
-    @Setter(AccessLevel.NONE)
-    private ProcessorProperties.Builder<ProcessorProperties> propertiesBuilder;
+    private final ProcessorProperties.Builder<ProcessorProperties> propertiesBuilder =
```
kawamuray, line 62: > avoid overwriting properties in SubscriptionBuilder - I can't see any good reason to do it

The intention is to provide a way to "clear" (reset) something set previously. So basically these methods taking collection of items replaces existing set rather than adding.
To keep this behavior consistent to other methods like `consumerConfig` and backward, I think we shouldn't change this. 
I'm happy to add another kind of method like `addProperties()` (`addProperty()`) which supports it and being more explicit tho.

testing/src/main/java/com/linecorp/decaton/testing/TestUtils.java
```diff
@@ -62,6 +72,7 @@ private static Properties defaultProducerProps(String bootstrapServers) {
     }
 
     public static final String DEFAULT_GROUP_ID = "test-group";
+    public static final Duration DEFINITELY_TOO_SLOW = Duration.ofSeconds(20);
```
kawamuray, line 75: > add timeouts of 20s in some places of the tests that were blocking forever, for easier debugging

curious why backtrace provided when test itself times out (`@Test(timeout = xxx)`) isn't sufficient to know at where it stuck?

kawamuray (changes requested):

testing/src/main/java/com/linecorp/decaton/testing/TestUtils.java
mauhiz: because it was not exactly stuck. Producer and Consumer were running and doing nothing.

processor/src/main/java/com/linecorp/decaton/processor/runtime/SubscriptionBuilder.java
mauhiz: I don't think that's what one would expect of a builder, but happy to oblige with your design constraints

processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
mauhiz: Ah, none. Indeed `SubscriptionMetrics` are only instantiated in `ProcessorSubscription` ctor. Removing

testing/src/main/java/com/linecorp/decaton/testing/TestUtils.java
```diff
@@ -189,20 +200,15 @@ public static void awaitCondition(String message,
...
             if (elapsedMillis >= timeoutMillis) {
                 throw new AssertionError(message);
             }
-            try {
-                Thread.sleep(100L);
-            } catch (InterruptedException e) {
-                Thread.currentThread().interrupt();
-                throw new RuntimeException(e);
-            }
+            LockSupport.parkNanos(TimeUnit.MILLISECONDS.toNanos(100));
```
kawamuray, line 211: What's the intention of this change? Seems like waiting just 100ms here isn't an actual problem?

kawamuray (approved):
LGTM

ocadaruma (approved):
Thank you!

LGTM except kawamuray's point.

testing/src/main/java/com/linecorp/decaton/testing/TestUtils.java
mauhiz: Just avoid having to deal with `InterruptedException`. No strong preference anyways

mauhiz:
Thanks for the reviews! Is the next release already scheduled?

kawamuray:
It should be in next few weeks, but if you're in hurry, I can make a release now.

mauhiz:
I'm not exactly in a hurry but finer-grained releases usually make happier users!

PR #133, opened 2021-11-25
Upgrade gradle to 7.3

### LINKED ISSUE #130: OutOfOrderCommitControlBenchmark does not compile
It got out of sync in #99 and for some reason the build did not detect the failure.

kawamuray:
I was trying to fix #130 and faced a strange error from shadow plugin, so attempted to upgrade, which led an effort to upgrade gradle itself. Well, it's a good timing tho, since we're far behind the latest.

Still haven't solved the problem, but I'll commit these gradle upgrade changes anyway.

Still performing some checks.

Might be considered as a breaking change because some transitive dependencies no longer provided to end-applications by default.


centraldogma/build.gradle
```diff
@@ -8,10 +8,11 @@ ext {
 }
 
 dependencies {
-    compile project(":processor")
+    implementation project(":processor")
```
ocadaruma, line 11: In my understanding, "dependencies of library-X which required to compile a project that uses library-X's public APIs" should be appeared in compile classpath (i.e. `api` dependency)". Is that correct?

refs: (https://docs.gradle.org/current/userguide/java_library_plugin.html)
> dependencies required to compile against the published library (i.e. API dependencies)

Then I think this should be `api`? Because this public API refers `processor`'s class. (`PropertyDefinition`)
https://github.com/line/decaton/blob/master/centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java#L105

I guess same applies to `centraldogma-client`. (And also for other modules. (e.g. `processor`'s `kafka-clients`, `micrometer` etc))

ocadaruma (commented):
Thanks!

Left some comments.

centraldogma/build.gradle
kawamuray: H-m. sounds like correct. let me fix that.

ocadaruma (approved):
Thanks, LGTM !

Issue #22, opened 2020-03-25
Add a feature to supply initial value for self CD registration


ocadaruma:
- When we create property file using `CentralDogmaPropertySupplier#register`, the file consists of default values.
- If we know appropriate value for a property at the beginning, this will require extra step to configure it
  * `CentralDogmaPropertySupplier#register`
  * update value on CentralDogma

fallaciesk:
Instead of letting `CentralDogmaPropertySupplier#register` create a JSON if missing with custom values, can we change it to NOT create a JSON at all ?

Say we construct subscription with multiple suppliers, and want to get different properties from different suppliers:
```
suppliers = [
    supplier1,  // property1
    supplier2,  //            property2            property4
    supplier3,  //                       property3
]
```

If we use `CentralDogmaPropertySupplier` and it automatically creates a JSON of all properties, with no matter default or custom values, then it shadows all suppliers after  it:

```
suppliers = [
    cd-supplier, // property1 default2 custom3 default4
    supplier2,   // (won't be used)
    supplier3,   // (won't be used)
]
```

To prevent this, we have to disable write permission at CD, and manually create a JSON with only wanted properties.
```
// JSON at CD
{
    "property1": "abc"
}

// code
suppliers = [cd-supplier, supplier2, supplier3]
```

If we change it to NOT automatically create JSON at all, then we can do like
```
suppliers = [
    cd-supplier,      // property1 
    supplier2,        //           property2
    default-supplier, // (all other properties with custom values)
]
```

Suppliers layered, and all missing values provided by a default supplier. And `supplier2` actually acts as the "custom values supplier".

lazmond3:
I’ll try this one.

kawamuray:
Solved by https://github.com/line/decaton/pull/134

PR #134, opened 2021-12-02
Add new method(overload) for CentralDogmaPropertySupplier#createPropertyFile/register

lazmond3:
# Relevant Issue
#22 

# Motivation
`CentralDogmaPropertySupplier#register` is used to create a property file with default settings.
The existing implementation of the function does not accept user-customized arguments.

If you know the appropriate values at the beginning and want to customize your settings, you need to take extra steps.
1. First, create a property file with `CentralDogmaPropertySupplier#register`.
2. Next, modify the values on CentralDogma.

This extra step is a bit annoying, so we want to let the function accept user-customized settings.

# Summary of Changes
- overload two new methods in `CentralDogmaPropertySupplier`
  1. `register` accepts `StaticPropertySupplier supplier`
      1. `register` converts `supplier` to `List<Property<?>>` for providing values to the next function: `createPropertyFile`.
  2. `createPropertyFile` accepts `List<Property<?>> properties`.

# Usage Example
```kotlin
CentralDogmaPropertySupplier.register(
    dogma,
    "sample-project",
    "app-using-decaton",
    "/decaton/settings.json",
    StaticPropertySupplier.of(
        Property.ofStatic(
            ProcessorProperties.CONFIG_PARTITION_CONCURRENCY,
            100
        ),
        Property.ofStatic(
            ProcessorProperties.CONFIG_MAX_PENDING_RECORDS,
            10000
        )
    )
)
```


centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
```diff
@@ -145,19 +149,43 @@ public static CentralDogmaPropertySupplier register(CentralDogma centralDogma, S
         return new CentralDogmaPropertySupplier(centralDogma, project, repository, filename);
     }
 
+    public static CentralDogmaPropertySupplier register(CentralDogma centralDogma, String project,
+                                                        String repository, String filename,
+                                                        StaticPropertySupplier supplier) {
```
kawamuray, line 154: I think in most cases people would use `StaticPropertySupplier` anyway, but we don't necessarily restrict this to be it, and just `PropertySupplier` (interface) is fine and more flexible?

kawamuray (changes requested):
Thanks for working on this! Left a minor comment WRT interface, but overall looks good 👍

lazmond3:
### Difference from the last time
- https://github.com/line/decaton/pull/134/commits/c17b3c7550e467a2857685f9f81cbe54d04c2f49 Changed the return type of `CentralDogmaPropertySupplier#defaultProperties()` from `JsonNode` to `List<Property<?>>`.
- https://github.com/line/decaton/pull/134/commits/df825c969b09c8a69639b5bfdcb5d2834f2ccd82 Added tests for `CentralDogmaPropertySupplier#register()`.

centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
```diff
@@ -145,19 +148,43 @@ public static CentralDogmaPropertySupplier register(CentralDogma centralDogma, S
         return new CentralDogmaPropertySupplier(centralDogma, project, repository, filename);
     }
 
+    public static CentralDogmaPropertySupplier register(CentralDogma centralDogma, String project,
+                                                        String repository, String filename,
+                                                        PropertySupplier supplier) {
+        List<Property<?>> properties = defaultProperties().stream().map(defaultProperty -> {
+            if (supplier.getProperty(defaultProperty.definition()).isPresent()) {
+                return supplier.getProperty(defaultProperty.definition()).get();
```
kawamuray, line 156: As a call of `getProperty` may involves I/O with external systems (like centraldogma), it is better to keep count of calling it as minimum as possible.
so let's do it like:

```java
prop = supplier.getProperty(...);
if (prop.ifPresent()) {
    return prop.get();
}
...
```

centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
```diff
@@ -145,19 +148,43 @@ public static CentralDogmaPropertySupplier register(CentralDogma centralDogma, S
...
+            } else {
+                return defaultProperty;
+            }
+        }).collect(Collectors.toList());
+
+        createPropertyFile(centralDogma, project, repository, filename, properties);
+        return new CentralDogmaPropertySupplier(centralDogma, project, repository, filename);
+    }
+
     private static void createPropertyFile(CentralDogma centralDogma, String project,
```
kawamuray, line 166: I think this overload no longer meaningful. Let's just use the below signature and pass default properties directly.

centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
```diff
@@ -225,12 +252,24 @@ private static long remainingTime(long totalTime, long startedTime) {
...
+                            property.definition().name(),
+                            objectMapper.valueToTree(property.value())
+                    );
+                }
+        );
+        return propertiesObjectNode;
+    }
+
+    // visible for testing
+    static List<Property<?>> defaultProperties() {
```
kawamuray, line 269: I think this functionality can be required by various supplier implementations? Maybe it's worth putting this method in `ProcessorProperties` rather than here privately.

centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
```diff
@@ -225,12 +252,24 @@ private static long remainingTime(long totalTime, long startedTime) {
...
+        List<Property<?>> properties = new ArrayList<>();
         ProcessorProperties.PROPERTY_DEFINITIONS
-                .forEach(definition -> properties.set(definition.name(),
-                                                      objectMapper.valueToTree(definition.defaultValue()))
-                );
+                .forEach(definition -> properties.add(new DynamicProperty(definition)));
```
kawamuray, line 272: There's no point to use `DynamicProperty` here. It's for modifiable property instance, so let's use `Property.ofStatic()` instead.

centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierTest.java
```diff
@@ -133,4 +144,106 @@ public void testGetPropertyAbsentName() {
...
+    public void testRegisterWithDefaultSettings() {
+        when(centralDogma.normalizeRevision(PROJECT_NAME, REPOSITORY_NAME, Revision.HEAD)).thenReturn(
+                CompletableFuture.completedFuture(Revision.HEAD)
+        );
+        when(centralDogma.listFiles(PROJECT_NAME, REPOSITORY_NAME, Revision.HEAD, FILENAME)).thenReturn(
+                CompletableFuture.completedFuture(Collections.emptyMap())
+        );
+        when(centralDogma.push(
+                PROJECT_NAME, REPOSITORY_NAME, Revision.HEAD,
+                String.format("[CentralDogmaPropertySupplier] Property file created: %s",
```
kawamuray, line 158: I don't think  we need to match this argument and it will just make shit test fragile for tiny change on commit message which is mostly pointless. So let's use type matcher here, like `any(String.class)`

centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierTest.java
```diff
@@ -133,4 +144,106 @@ public void testGetPropertyAbsentName() {
...
+                Change.ofJsonUpsert(FILENAME, defaultProperties()))
+        ).thenReturn(
+                CompletableFuture.completedFuture(
+                        new PushResult(Revision.HEAD, 1)
+                )
+        );
+
+        CentralDogmaPropertySupplier.register(centralDogma, PROJECT_NAME, REPOSITORY_NAME, FILENAME);
+
+        verify(centralDogma, times(1)).normalizeRevision(PROJECT_NAME, REPOSITORY_NAME, Revision.HEAD);
```
kawamuray, line 169: I think what we wanna test here is, if calling `register()` actually tries to create an expected file with expected content, so the only check against `push()` is meaningful to be testing subject?

centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierTest.java
```diff
@@ -133,4 +144,106 @@ public void testGetPropertyAbsentName() {
...
+        when(centralDogma.normalizeRevision(PROJECT_NAME, REPOSITORY_NAME, Revision.HEAD)).thenReturn(
+                CompletableFuture.completedFuture(Revision.HEAD)
+        );
+        when(centralDogma.listFiles(PROJECT_NAME, REPOSITORY_NAME, Revision.HEAD, FILENAME)).thenReturn(
+                CompletableFuture.completedFuture(Collections.emptyMap())
+        );
+
+        when(centralDogma.push(
+                PROJECT_NAME, REPOSITORY_NAME, Revision.HEAD,
+                String.format("[CentralDogmaPropertySupplier] Property file created: %s", FILENAME),
```
kawamuray, line 226: here too.

centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierTest.java
```diff
@@ -133,4 +144,106 @@ public void testGetPropertyAbsentName() {
...
+        );
+
+        CentralDogmaPropertySupplier.register(centralDogma, PROJECT_NAME, REPOSITORY_NAME, FILENAME,
+                                              properties);
+
+        verify(centralDogma, times(1)).normalizeRevision(PROJECT_NAME, REPOSITORY_NAME, Revision.HEAD);
+        verify(centralDogma, times(1)).listFiles(PROJECT_NAME, REPOSITORY_NAME, Revision.HEAD, FILENAME);
+        verify(centralDogma, times(1))
+                .push(PROJECT_NAME, REPOSITORY_NAME, Revision.HEAD,
+                      String.format("[CentralDogmaPropertySupplier] Property file created: %s", FILENAME),
```
kawamuray, line 241: here too.

centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierTest.java
```diff
@@ -133,4 +144,106 @@ public void testGetPropertyAbsentName() {
...
+                                            String.format(
+                                                    "[CentralDogmaPropertySupplier] Property file created: %s",
+                                                    FILENAME),
+                                            Change.ofJsonUpsert(FILENAME, defaultProperties()));
+
+    }
+
+    @Test
+    public void testRegisterWithCustomizedSettings() {
+        final Integer settingForPartitionConcurrency = 188;
```
kawamuray, line 181: Let's stuck with primitive types when there's no point to use a boxed type. `int`

kawamuray (changes requested):
Thanks for your PR! Left some comments.

centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierTest.java
lazmond3: I changed the variable type to `int` from `Integer`. https://github.com/line/decaton/pull/134/commits/382f70617ba6fdfb9659f43fc69dd915a89d0c27

centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
```diff
@@ -145,19 +148,43 @@ public static CentralDogmaPropertySupplier register(CentralDogma centralDogma, S
         return new CentralDogmaPropertySupplier(centralDogma, project, repository, filename);
     }
 
+    public static CentralDogmaPropertySupplier register(CentralDogma centralDogma, String project,
```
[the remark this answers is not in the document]
lazmond3: I added a javadoc for the new added `register` method: https://github.com/line/decaton/pull/134/commits/e8c311bf4d58d4f3b2ba5a9ec33519ae6a2ff756

centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
lazmond3: I use a temporary variable for keeping the value in external systems. https://github.com/line/decaton/pull/134/commits/8d38efefa231e638194e922b2f1a37ded6641185

centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
lazmond3: I deleted this original method `createPropertyFile` (no longer used), since a new method which has more arguments was added. https://github.com/line/decaton/pull/134/commits/6b49b94c9a1f6cb7493fb3f13be8d4240f4c1dff

centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
lazmond3: I moved this method `defaultProperties`, which returns `List<Property<?>>`,  to `ProcessorProperties`. https://github.com/line/decaton/pull/134/commits/03ef312262463e153abaf00dbd83410702f4de2b

centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
lazmond3: https://github.com/line/decaton/pull/134/commits/6d9e69e9a99e3e52ca4ece5f2c1629dcce846497

centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierTest.java
lazmond3: I changed the code to use mockito `any(String.class)`, since a central dogma's commit message is not target of test for this PR's functionality.
https://github.com/line/decaton/pull/134/commits/f2ee11da05f8db02995aba1f609d0dd2271d680b

centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierTest.java
lazmond3: I deleted the verifying lines: https://github.com/line/decaton/pull/134/commits/61cea291f0dc7c04ef0885a1dea607fb71959819

centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierTest.java
lazmond3: https://github.com/line/decaton/commit/f2ee11da05f8db02995aba1f609d0dd2271d680b

centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierTest.java
lazmond3: https://github.com/line/decaton/commit/f2ee11da05f8db02995aba1f609d0dd2271d680b

centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
```diff
@@ -141,23 +144,51 @@ public void close() {
...
+    }
+
+    /**
+     * Create a default property file if it doesn't exist on Central Dogma and
+     * return a {@link CentralDogmaPropertySupplier}.
+     * @param centralDogma a {@link CentralDogma} instance to use to access Central Dogma server.
+     * @param project the project name where the properties are placed.
+     * @param repository the repository name where the properties are placed.
+     * @param filename the name of the file containing properties as top-level fields.
+     * @param supplier user-customized settings for kafka-consumer.
```
kawamuray, line 158: These properties are of decaton, not kafka-consumer. So let's reword it like: `a {@link PropertySupplier} which provides set of properties that wants to have a custom initial value`

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorProperties.java
```diff
@@ -218,4 +219,10 @@
     public ProcessorProperties(Map<PropertyDefinition<?>, Property<?>> properties) {
         super(properties);
     }
+
+    public static List<Property<?>> defaultProperties() {
```
kawamuray, line 223: Please give it even a short javadoc since its also a public interface.

centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierTest.java
kawamuray: Hmm, looks like this part and one in below hasn't changed yet?

kawamuray (changes requested):
Thanks for follow-up. Just few more things.

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorProperties.java
lazmond3: I added a short description for this method: 2e70198

centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierTest.java
lazmond3: I'm sorry, I failed to update the code exhaustively.
I changed the `when` parts to use `any`: 0f2db81

centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
lazmond3: I reworded the javadoc for supplier: 
`a {@link PropertySupplier} which provides a set of properties with customized initial values`
https://github.com/line/decaton/pull/134/commits/2e70198681ebd07786d1e4e69004364e51b31f30

kawamuray (approved):
LGTM, thanks!

kawamuray:
test is failing.

centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierTest.java
```diff
@@ -133,4 +145,111 @@ public void testGetPropertyAbsentName() {
...
+
+    }
+
+    @Test
+    public void testRegisterWithCustomizedSettings() {
+        final int settingForPartitionConcurrency = 188;
+        final int settingForMaxPendingRecords = 121212;
+        final int whenCentralDogmaPushed = 111111;
+
+        List<Property<?>> properties = ProcessorProperties
```
kawamuray, line 186: I don't see why we need full list of proeprties here, not only custom ones but also these with defaults?
Since the spec of newly added `register` API is such that "Creates a node with default properties, but you can specify some prop's initial value optionally", the realistic usage will look like:

```java
CentralDogmaPropertySupplier.register(..., StaticPropertySupplier.of(
    Property.ofStatic(...)
));
```

so what we should test here is passing just partial list of properties that we want to customize? such as:


```java
    @Test
    public void testRegisterWithCustomizedSettings() {
        final int settingForPartitionConcurrency = 188;
        final int settingForMaxPendingRecords = 121212;
        final int whenCentralDogmaPushed = 111111;

        final PropertySupplier supplier = StaticPropertySupplier.of(
                Property.ofStatic(ProcessorProperties.CONFIG_PARTITION_CONCURRENCY,
                                  settingForPartitionConcurrency),
                Property.ofStatic(ProcessorProperties.CONFIG_MAX_PENDING_RECORDS,
                                  settingForMaxPendingRecords)
        );
        final List<Property<?>> listProperties = ProcessorProperties
                .defaultProperties()
                .stream()
                .map(defaultProperty -> {
                    Optional<? extends Property<?>> prop = supplier.getProperty(defaultProperty.definition());
                    if (prop.isPresent()) {
                        return prop.get();
                    } else {
                        return defaultProperty;
                    }
                }).collect(Collectors.toList());

...
```

kawamuray (changes requested):

centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierTest.java
lazmond3: I agree with you. I don't need to provide all of the properties if I don't change the value of it.
However, on the other hand I have to build a full list of properties for verifying mock.

Changed: providing only customized properties
e6ae753

kawamuray (approved):
LGTM, once again :)

PR #136, opened 2021-12-06
shadeAllDependencies needs to be evaluated after subproject configuation

kawamuray:
Follow up for #133


ocadaruma (approved):
Thanks, nice catch.

Confirmed `debm.sh` works on local.

PR #138, opened 2021-12-06
Fixup; using out_dir as tmp filename doesn't work

kawamuray:
Fixup for #137

### DIFF
--- a/cb/run-bm.sh
+++ b/cb/run-bm.sh
@@ -16,7 +16,7 @@ $root_dir/gradlew --no-daemon clean benchmark:shadowJar
 
 function run_with_opts() {
     name=$1; shift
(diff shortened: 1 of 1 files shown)


cb/run-bm.sh
```diff
@@ -16,7 +16,7 @@ $root_dir/gradlew --no-daemon clean benchmark:shadowJar
 
 function run_with_opts() {
     name=$1; shift
-    tmp="/tmp/${out_dir}-${name}-benchmark.json"
+    tmp="/tmp/decaton-cb-$$.json"
```
ocadaruma, line 19: Maybe `mktemp` is more simpler?

ocadaruma (approved):
LGTM

Issue #128, opened 2021-11-05
Provide BatchingProcessor as a built-in processor


ocadaruma:
- Task batching is a common-pattern that many Decaton users often implement by their own
  * i.e. Batching several tasks of type `T` to `List<T>` and process them at once. e.g. when downstream-DB supports batching I/O (which often very efficient)
  * Batch-flushing should be done in time-based and size-based.
- So it's better to provide BatchingProcessor built-in to meet the common needs

ryamagishi:
I'll try this issue.

ocadaruma:
Thank you!
I assigned you.

ocadaruma:
fixed by #139, #150

PR #139, opened 2021-12-09
Add batching processor

ryamagishi:
# Motivation
- Task batching is a common-pattern that many Decaton users often implement by their own.
   - i.e. Batching several tasks of type T to List<T> and process them at once. e.g. when downstream-DB supports batching I/O (which often very efficient)
   - Batch-flushing should be done in time-based and size-based.
- So it's better to provide BatchingProcessor built-in to meet the common needs
 
Related Issue: https://github.com/line/decaton/issues/128

### DIFF
--- a/processor/src/it/java/com/linecorp/decaton/processor/BatchingProcessorTest.java
+++ b/processor/src/it/java/com/linecorp/decaton/processor/BatchingProcessorTest.java
@@ -64,32 +64,4 @@ public class BatchingProcessorTest {
             .build()
             .run();
     }
-
-    // TODO: Fix test or test's target. This test does not succeed.
-//    @Test(timeout = 30000)
-//    public void testBatchingProcessor_capacity() throws Exception {
-//        Random rand = randomRule.random();
-//        ProcessorTestSuite
-//            .builder(rule)
-//            .configureProcessorsBuilder(builder -> builder.thenProcess(
-//                new BatchingProcessor<TestTask>(Long.MAX_VALUE, 100) {
-//                    @Override
-//                    protected void processBatchingTasks(List<BatchingTask<TestTask>> batchingTasks) {
-//                        // adding some random delay to simulate realistic usage
-//                        try {
-//                            Thread.sleep(rand.nextInt(10));
-//                        } catch (InterruptedException e) {
-//                            Thread.currentThread().interrupt();
-//                            throw new RuntimeException(e);
-//                        }
-//                        batchingTasks.forEach(batchingTask -> batchingTask.completion().complete());
-//                    }
-//                }
-//            ))
-//            .propertySupplier(StaticPropertySupplier.of(
-//                Property.ofStatic(ProcessorProperties.CONFIG_PARTITION_CONCURRENCY, 16)
-//            ))
-//            .build()
-//            .run();
-//    }
 }
--- a/processor/src/it/java/com/linecorp/decaton/processor/BatchingProcessorTest.java
+++ b/processor/src/it/java/com/linecorp/decaton/processor/BatchingProcessorTest.java
@@ -39,12 +39,12 @@ public class BatchingProcessorTest {
     public RandomRule randomRule = new RandomRule();
 
     @Test(timeout = 30000)
-    public void testBatchingProcessor_lingerMillis() throws Exception {
+    public void testBatchingProcessor() throws Exception {
         Random rand = randomRule.random();
         ProcessorTestSuite
             .builder(rule)
             .configureProcessorsBuilder(builder -> builder.thenProcess(
-                new BatchingProcessor<TestTask>(1000, Integer.MAX_VALUE) {
+                new BatchingProcessor<TestTask>(1000, 100) {
                     @Override
                     protected void processBatchingTasks(List<BatchingTask<TestTask>> batchingTasks) {
                         // adding some random delay to simulate realistic usage
--- a/processor/src/it/java/com/linecorp/decaton/processor/BatchingProcessorTest.java
+++ b/processor/src/it/java/com/linecorp/decaton/processor/BatchingProcessorTest.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2022 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.decaton.processor;
+
+import java.util.List;
+import java.util.Random;
+
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.Test;
+
+import com.linecorp.decaton.processor.processors.BatchingProcessor;
+import com.linecorp.decaton.processor.runtime.ProcessorProperties;
+import com.linecorp.decaton.processor.runtime.Property;
+import com.linecorp.decaton.processor.runtime.StaticPropertySupplier;
+import com.linecorp.decaton.testing.KafkaClusterRule;
+import com.linecorp.decaton.testing.RandomRule;
+import com.linecorp.decaton.testing.processor.ProcessorTestSuite;
+import com.linecorp.decaton.testing.processor.TestTask;
+
+public class BatchingProcessorTest {
+    @ClassRule
+    public static KafkaClusterRule rule = new KafkaClusterRule();
+    @Rule
+    public RandomRule randomRule = new RandomRule();
+
+    @Test(timeout = 30000)
+    public void testBatchingProcessor_lingerMillis() {
+        Random rand = randomRule.random();
+        ProcessorTestSuite
+            .builder(rule)
+            .configureProcessorsBuilder(builder -> builder.thenProcess(
+                new BatchingProcessor<TestTask>(1000, Integer.MAX_VALUE) {
+                    @Override
+                    protected void processBatchingTasks(List<BatchingTask<TestTask>> batchingTasks) {
+                        // adding some random delay to simulate realistic usage
+                        try {
+                            Thread.sleep(rand.nextInt(10));
+                        } catch (InterruptedException e) {
+                            Thread.currentThread().interrupt();
+                            throw new RuntimeException(e);
+                        }
+                        batchingTasks.forEach(batchingTask -> batchingTask.completion().complete());
+                    }
+                }
+            ))
+            .propertySupplier(StaticPropertySupplier.of(
+                Property.ofStatic(ProcessorProperties.CONFIG_PARTITION_CONCURRENCY, 16)
+            ))
+            .build()
+            .run();
+    }
+
+    // TODO: Fix test or test's target. This test does not succeed.
+//    @Test(timeout = 30000)
+//    public void testBatchingProcessor_capacity() {
+//        Random rand = randomRule.random();
+//        ProcessorTestSuite
+//            .builder(rule)
+//            .configureProcessorsBuilder(builder -> builder.thenProcess(
+//                new BatchingProcessor<TestTask>(Long.MAX_VALUE, 100) {
+//                    @Override
+//                    protected void processBatchingTasks(List<BatchingTask<TestTask>> batchingTasks) {
+//                        // adding some random delay to simulate realistic usage
+//                        try {
+//                            Thread.sleep(rand.nextInt(10));
+//                        } catch (InterruptedException e) {
+//                            Thread.currentThread().interrupt();
+//                            throw new RuntimeException(e);
+//                        }
+//                        batchingTasks.forEach(batchingTask -> batchingTask.completion().complete());
+//                    }
+//                }
+//            ))
+//            .propertySupplier(StaticPropertySupplier.of(
+//                Property.ofStatic(ProcessorProperties.CONFIG_PARTITION_CONCURRENCY, 16)
+//            ))
+//            .build()
+//            .run();
+//    }
+}
--- a/processor/src/it/java/com/linecorp/decaton/processor/BatchingProcessorTest.java
+++ b/processor/src/it/java/com/linecorp/decaton/processor/BatchingProcessorTest.java
@@ -39,7 +39,7 @@ public class BatchingProcessorTest {
     public RandomRule randomRule = new RandomRule();
 
     @Test(timeout = 30000)
-    public void testBatchingProcessor_lingerMillis() {
+    public void testBatchingProcessor_lingerMillis() throws Exception {
         Random rand = randomRule.random();
         ProcessorTestSuite
             .builder(rule)
@@ -67,7 +67,7 @@ public class BatchingProcessorTest {
 
     // TODO: Fix test or test's target. This test does not succeed.
 //    @Test(timeout = 30000)
-//    public void testBatchingProcessor_capacity() {
+//    public void testBatchingProcessor_capacity() throws Exception {
 //        Random rand = randomRule.random();
 //        ProcessorTestSuite
 //            .builder(rule)
--- a/processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
@@ -68,7 +68,7 @@ public abstract class BatchingProcessor<T> implements DecatonProcessor<T> {
      * @param capacity size limit for this processor. Every time tasks’size reaches capacity,
      * tasks in past before reaching capacity are pushed to {@link BatchingTask#processBatchingTasks(List)}.
      */
-    public BatchingProcessor(long lingerMillis, int capacity) {
+    protected BatchingProcessor(long lingerMillis, int capacity) {
         this.lingerMillis = lingerMillis;
         this.capacity = capacity;
 
--- a/processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
@@ -17,7 +17,6 @@
 package com.linecorp.decaton.processor.processors;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
@@ -40,7 +39,7 @@ import lombok.experimental.Accessors;
 public abstract class BatchingProcessor<T> implements DecatonProcessor<T> {
 
     private final ScheduledExecutorService executor;
-    private final List<BatchingTask<T>> windowedTasks = Collections.synchronizedList(new ArrayList<>());
+    private List<BatchingTask<T>> currentBatch = new ArrayList<>();
     private final long lingerMillis;
     private final int capacity;
 
@@ -95,38 +94,36 @@ public abstract class BatchingProcessor<T> implements DecatonProcessor<T> {
         scheduleFlush();
     }
 
-    private void flush(List<BatchingTask<T>> batchingTasks) {
-        if (batchingTasks.isEmpty()) {
-            return;
-        }
-        processBatchingTasks(batchingTasks);
-    }
-
-    private Runnable periodicallyFlushTask() {
-        return () -> {
-            synchronized (windowedTasks) {
-                flush(new ArrayList<>(windowedTasks));
-                windowedTasks.clear();
+    private void periodicallyFlushTask() {
+        final List<BatchingTask<T>> batch;
+        synchronized (this) {
+            if (!currentBatch.isEmpty()) {
+                batch = currentBatch;
+                currentBatch = new ArrayList<>();
+            } else {
+                batch = null;
             }
-            scheduleFlush();
-        };
+        }
+        if (batch != null) {
+            processBatchingTasks(batch);
+        }
+        scheduleFlush();
     }
 
     private void scheduleFlush() {
-        executor.schedule(periodicallyFlushTask(), lingerMillis, TimeUnit.MILLISECONDS);
+        executor.schedule(this::periodicallyFlushTask, lingerMillis, TimeUnit.MILLISECONDS);
     }
 
     @Override
     public void process(ProcessingContext<T> context, T task) throws InterruptedException {
-        BatchingTask<T> newTask = new BatchingTask<>(context.deferCompletion(), context, task);
-        windowedTasks.add(newTask);
-        if (windowedTasks.size() >= this.capacity) {
-            synchronized (windowedTasks) {
-                List<BatchingTask<T>> currentBatchingTasks = new ArrayList<>(windowedTasks);
-                flush(currentBatchingTasks.subList(0, this.capacity));
-                windowedTasks.clear();
-                windowedTasks.addAll(currentBatchingTasks.subList(this.capacity, currentBatchingTasks.size()));
+        synchronized (this) {
+            if (currentBatch.size() >= this.capacity) {
+                final List<BatchingTask<T>> batch = currentBatch;
+                executor.submit(() -> processBatchingTasks(batch));
+                currentBatch = new ArrayList<>();
             }
+            BatchingTask<T> newTask = new BatchingTask<>(context.deferCompletion(), context, task);
+            currentBatch.add(newTask);
         }
     }
 
--- a/processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
@@ -27,6 +27,7 @@ import com.linecorp.decaton.processor.DecatonProcessor;
 import com.linecorp.decaton.processor.DeferredCompletion;
 import com.linecorp.decaton.processor.ProcessingContext;
 import com.linecorp.decaton.processor.Completion;
+import com.linecorp.decaton.processor.runtime.internal.Utils;
 
 import lombok.RequiredArgsConstructor;
 import lombok.Value;
@@ -66,12 +67,12 @@ public abstract class BatchingProcessor<T> implements DecatonProcessor<T> {
         this.lingerMillis = lingerMillis;
         this.capacity = capacity;
 
-        ScheduledThreadPoolExecutor scheduledExecutor = new ScheduledThreadPoolExecutor(1, r -> {
-            Thread th = new Thread(r);
-            th.setName("Decaton" + BatchingProcessor.class.getSimpleName()
-                       + '/' + System.identityHashCode(this));
-            return th;
-        });
+        ScheduledThreadPoolExecutor scheduledExecutor = new ScheduledThreadPoolExecutor(
+            1,
+            Utils.namedThreadFactory(
+                "Decaton" + BatchingProcessor.class.getSimpleName() + '/' + System.identityHashCode(this)
+            )
+        );
 
         // For faster shutdown cancel all pending flush on executor shutdown.
         // In fact for this purpose we have two options here:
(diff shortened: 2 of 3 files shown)


processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    private void scheduleFlush() {
+        executor.schedule(flushTask(), lingerMillis, TimeUnit.MILLISECONDS);
+    }
+
+    @Override
+    public void process(ProcessingContext<T> context, T task) throws InterruptedException {
+        BatchingTask<T> newTask = new BatchingTask<>(context.deferCompletion(), context, task);
+        boolean isInitialTask = windowedTasks.isEmpty();
```
ryamagishi, line 116: The motivation is to make a decision before adding a Task, and to scheduleFlush after adding a Task.

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
```diff
@@ -0,0 +1,137 @@
...
+
+    @Override
+    public void close() throws Exception {
+        executor.shutdown();
+        executor.awaitTermination(Long.MAX_VALUE, TimeUnit.SECONDS);
+    }
+
+    /**
+     * *MUST* call {@link BatchingTask#completion}'s {@link DeferredCompletion#complete()} or
+     * {@link BatchingTask#context}'s {@link ProcessingContext#retry()} method.
```
ryamagishi, line 134: As a result of worrying about the retry process, I left it to the implementation side, but I think that it should be improved...

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
```diff
@@ -0,0 +1,137 @@
...
+    public BatchingProcessor(long lingerMillis, int capacity) {
+        this(lingerMillis, capacity, null);
+    }
+
+    private void flush() {
+        if (!windowedTasks.isEmpty()) {
+            return;
+        }
+        synchronized (windowedTasks) {
+            processBatchingTasks(windowedTasks);
```
ocadaruma, line 100: Seems like clearing passed batchingTasks is also implementer's responsibility?
I guess it's not good design, as it could be error-prone.

Let's pass copy of windowedTasks and clear it?

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
```diff
@@ -0,0 +1,137 @@
...
+    public BatchingProcessor(long lingerMillis) {
+        this(lingerMillis, Integer.MAX_VALUE, null);
+    }
+
+    public BatchingProcessor(long lingerMillis, int capacity) {
+        this(lingerMillis, capacity, null);
+    }
+
+    private void flush() {
+        if (!windowedTasks.isEmpty()) {
```
ocadaruma, line 96: Hm, is that intentional? (not `windowedTasks.isEmpty()`?)

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
```diff
@@ -0,0 +1,137 @@
...
+            return;
+        }
+        synchronized (windowedTasks) {
+            processBatchingTasks(windowedTasks);
+        }
+    }
+
+    // visible for testing
+    Runnable flushTask() {
+        return this::flush;
```
ocadaruma, line 106: Seems flushTask() isn't called periodically?

ocadaruma (commented):
Thanks for submitting a patch!
Added few early feedbacks.

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
ryamagishi: Thank you for the advice! I fixed it.
1a2e2c2

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
ryamagishi: Sorry... I made a mistake.
4fdf913

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
ryamagishi: I was thinking about the following, but it was certainly difficult to understand, so I fixed it!
1. When the first task comes in, `scheduleFlush()` is called.
2. `flush()` empties windowsTasks.
3. When the next task comes in after `flush()`,  `scheduleFlush()` is called because windowTasks is empty.
2ef1bc0

ryamagishi:
I'm sorry for the late response 🙇 
I fixed the github commented part, added comments and tests for `BatchingProcessor` and changed PR status from draft.
I would be grateful if you could review this when you have time.

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
```diff
@@ -0,0 +1,138 @@
...
+    private void scheduleFlush() {
+        executor.schedule(periodicallyFlushTask(), lingerMillis, TimeUnit.MILLISECONDS);
+    }
+
+    @Override
+    public void process(ProcessingContext<T> context, T task) throws InterruptedException {
+        BatchingTask<T> newTask = new BatchingTask<>(context.deferCompletion(), context, task);
+        windowedTasks.add(newTask);
+        if (windowedTasks.size() >= this.capacity) {
+            flush();
```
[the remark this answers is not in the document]
ocadaruma: Hm, then the problem is at the time `flush()` called, windowedTasks may contain more tasks than batch size, it's also not desirable.

Ok, then what do you think about following strategy?
- Instead of using single final `windowedTasks`, have non-final `List<BatchingTask>` for "current" batch.
- When "current" batch exceeds batch size, submit it to flusher and substitute current batch with new List.

Decaton will pause consumption if there's too many pending tasks, so even with this strategy, heap usage will not grow indefinitely.

processor/src/test/java/com/linecorp/decaton/processor/processors/BatchingProcessorTest.java
```diff
@@ -0,0 +1,117 @@
...
+        BatchingProcessor<HelloTask> processor = buildProcessor(lingerMs, Integer.MAX_VALUE);
+
+        HelloTask task1 = buildHelloTask("one", 1);
+        HelloTask task2 = buildHelloTask("two", 2);
+        HelloTask task3 = buildHelloTask("three", 3);
+        HelloTask task4 = buildHelloTask("four", 4);
+
+        processor.process(context, task1);
+        processor.process(context, task2);
+        Thread.sleep(lingerMs * 2); // doubling just to make sure flush completed in background
```
ocadaruma, line 85: Using Thread.sleep to wait something happen makes test result flaky.
Let's use CountDownLatch to make sure the condition happens. (Referring CompactionProcessorTest may be helpful)

ocadaruma (commented):
Sorry for delaying the review!

Reviewed 1st round. PTAL comments.

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
```diff
@@ -0,0 +1,138 @@
...
+import lombok.Getter;
+import lombok.experimental.Accessors;
+
+/**
+ * An Abstract {@link DecatonProcessor} to Batch several tasks of type T to List<T> and process them at once.
+ * e.g. when downstream-DB supports batching I/O (which often very efficient).
+ * Batch-flushing should be done in time-based and size-based.
+ * @param <T> type of task to batch
+ */
+abstract public class BatchingProcessor<T> implements DecatonProcessor<T> {
```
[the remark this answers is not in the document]
ryamagishi: I fixed it. 4471e71

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Instantiate {@link BatchingProcessor}.
+     * If you only need one limit, please set large enough value to another.
+     * @param lingerMillis time limit for this processor. On every lingerMillis milliseconds,
+     * tasks in past lingerMillis milliseconds are pushed to {@link BatchingTask#processBatchingTasks(List)}.
+     * @param capacity size limit for this processor. Every time tasks’size reaches capacity,
+     * tasks in past before reaching capacity are pushed to {@link BatchingTask#processBatchingTasks(List)}.
+     */
+    public BatchingProcessor(long lingerMillis, int capacity) {
```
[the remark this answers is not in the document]
ryamagishi: I fixed it. f6f97ba

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
```diff
@@ -0,0 +1,138 @@
...
+    private final ScheduledExecutorService executor;
+    private final List<BatchingTask<T>> windowedTasks = Collections.synchronizedList(new ArrayList<>());
+    private final long lingerMillis;
+    private final int capacity;
+
+    @Getter
+    @Accessors(fluent = true)
+    public static class BatchingTask<T> {
+        @Getter(AccessLevel.NONE)
+        public final Completion completion;
```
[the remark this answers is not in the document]
ryamagishi: I fixed it. c668bb3

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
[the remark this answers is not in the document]
ryamagishi: I'm sorry, I probably don't understand what you're saying...
I tried to correct it with my own understanding, but please point out if it is wrong 🙇 
cb78a8b

processor/src/test/java/com/linecorp/decaton/processor/processors/BatchingProcessorTest.java
ryamagishi: Thank you for your reference.
I fixed it so that the test result is stable as much as possible.
357f2fb

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
```diff
@@ -0,0 +1,145 @@
...
+    public void close() throws Exception {
+        executor.shutdown();
+        executor.awaitTermination(Long.MAX_VALUE, TimeUnit.SECONDS);
+    }
+
+    /**
+     * *MUST* call {@link BatchingTask#completion}'s {@link DeferredCompletion#complete()} or
+     * {@link BatchingTask#context}'s {@link ProcessingContext#retry()} method.
+     */
+    abstract void processBatchingTasks(List<BatchingTask<T>> batchingTasks);
```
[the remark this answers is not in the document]
ryamagishi: I fixed it. 5ae0959

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
```diff
@@ -0,0 +1,145 @@
...
+    // visible for testing
+    protected void scheduleFlush() {
+        executor.schedule(periodicallyFlushTask(), lingerMillis, TimeUnit.MILLISECONDS);
+    }
+
+    @Override
+    public void process(ProcessingContext<T> context, T task) throws InterruptedException {
+        BatchingTask<T> newTask = new BatchingTask<>(context.deferCompletion(), context, task);
+        windowedTasks.add(newTask);
+        if (windowedTasks.size() >= this.capacity) {
```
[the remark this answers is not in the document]
ryamagishi: >I think this should be executed in scheduler thread to avoid unnecessary lock contention between Decaton's processor thread and scheduler thread

Thank you! I finally understood the meaning of the above sentence.
`executor.submit(() -> processBatchingTasks(batch));`  is important.
I fixed it. 9e35f1b

ryamagishi:
https://github.com/line/decaton/pull/139#commitcomment-63863727
Thanks for your test code advice.
While trying, I noticed that the abstract class mock wasn't working as expected.
So I fixed it to write test code simply without mocking. 
ref: https://www.baeldung.com/junit-test-abstract-class

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
```diff
@@ -0,0 +1,141 @@
...
+    @Accessors(fluent = true)
+    public static class BatchingTask<T> {
+        @Getter
+        private final Completion completion;
+        @Getter
+        private final ProcessingContext<T> context;
+        @Getter
+        private final T task;
+
+        private BatchingTask(Completion completion, ProcessingContext<T> context, T task) {
```
ocadaruma, line 56: Yeah basically this class is expected to be instantiated only from BatchingProcessor though, I think sometimes users may want to instantiate this class when they write unite-test for their processor that implements BatchingProcessor.

So let's make the constructor public? (then, just use `@RequiredArgsConstructor`) might be simple?

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
```diff
@@ -0,0 +1,141 @@
...
+    private final int capacity;
+
+    @Getter
+    @Accessors(fluent = true)
+    public static class BatchingTask<T> {
+        @Getter
+        private final Completion completion;
+        @Getter
+        private final ProcessingContext<T> context;
+        @Getter
```
ocadaruma, line 53: `@Getter` is used on class-level so field-level `@Getter` is not necessary I think

processor/src/test/java/com/linecorp/decaton/processor/processors/BatchingProcessorTest.java
```diff
@@ -0,0 +1,119 @@
...
+    @Rule
+    public MockitoRule rule = MockitoJUnit.rule();
+
+    @Mock
+    private ProcessingContext<HelloTask> context;
+
+    @Mock
+    private Completion completion;
+
+    private final List<HelloTask> processedTasks = new ArrayList<>();
```
ocadaruma, line 51: Maybe this should be wrapped with `Collections.synchronizedList` ? Because this will be read from test thread while mutated by BatchingProcessor's executor thread.

ocadaruma (commented):
Sorry for delaying the review...

Left few minor comments, but almost LGTM!

Besides, would you mind to do few more works?:
- Adding an integration test to see Decaton's core functionality (`at_least_once, process_ordering, serial_processing`) is preserved even on BatchingProcessor?
  * like this: https://github.com/line/decaton/blob/master/processor/src/it/java/com/linecorp/decaton/processor/CoreFunctionalityTest.java#L52-L64
- (Fine to be done in separated PR) adding a documentation like this? https://github.com/line/decaton/blob/master/docs/task-compaction.adoc

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
```diff
@@ -0,0 +1,141 @@
...
+import com.linecorp.decaton.processor.ProcessingContext;
+import com.linecorp.decaton.processor.Completion;
+
+import lombok.Getter;
+import lombok.experimental.Accessors;
+
+/**
+ * An Abstract {@link DecatonProcessor} to Batch several tasks of type {@code T} to {@code List<T>}
+ * and process them at once. e.g. when downstream-DB supports batching I/O (which often very efficient).
+ * Batch-flushing should be done in time-based and size-based.
```
kawamuray, line 36: `should be done` -> `is done`?

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
```diff
@@ -0,0 +1,141 @@
...
+     * tasks in past lingerMillis milliseconds are pushed to {@link BatchingTask#processBatchingTasks(List)}.
+     * @param capacity size limit for this processor. Every time tasks’size reaches capacity,
+     * tasks in past before reaching capacity are pushed to {@link BatchingTask#processBatchingTasks(List)}.
+     */
+    protected BatchingProcessor(long lingerMillis, int capacity) {
+        this.lingerMillis = lingerMillis;
+        this.capacity = capacity;
+
+        ScheduledThreadPoolExecutor scheduledExecutor = new ScheduledThreadPoolExecutor(1, r -> {
+            Thread th = new Thread(r);
```
kawamuray, line 76: I think you can use `Utils.namedThreadFactory` for this.

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
```diff
@@ -0,0 +1,141 @@
...
+ * @param <T> type of task to batch
+ */
+public abstract class BatchingProcessor<T> implements DecatonProcessor<T> {
+
+    private final ScheduledExecutorService executor;
+    private List<BatchingTask<T>> currentBatch = new ArrayList<>();
+    private final long lingerMillis;
+    private final int capacity;
+
+    @Getter
```
kawamuray, line 46: can't this be simply a `@Value`?

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
kawamuray: Since this is the main method of this class that we expect users to implement, the document should be bit more informative IMO. 

At least it should cover topics like what to do after complete processing batch of tasks. (call `.completion().complete()` or  `retry()` for each right?),

what happens if an error is thrown, which threads may possibly call this method etc.

kawamuray (changes requested):
Reviewed mainly for interfaces, looks mostly good! Just few points to suggest.

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
```diff
@@ -0,0 +1,141 @@
...
+                batch = currentBatch;
+                currentBatch = new ArrayList<>();
+            } else {
+                batch = null;
+            }
+        }
+        if (batch != null) {
+            processBatchingTasks(batch);
+        }
+        scheduleFlush();
```
ocadaruma, line 110: H-m, I realized that unlimited number of scheduleFlush task may be queued in executor when `processBatchingTasks` for head-batch is somehow blocked.
Also, I think scheduleFlush task may be queued even when there are no incoming tasks, it sounds wasteful. (then the flush task may return immediately so not gonna be big problem though)

So I guess we need some fixes like:
- Schedule flushTask only when new batch is created
- Don't schedule further flushTask if there's pending one

ocadaruma: > unlimited number of scheduleFlush task may be queued

This sounds not true because new delayed task will not be queued until this task returns.
Please nevermind my comment.

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
ryamagishi: I fixed it. 3475850

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
ryamagishi: I fixed it. 3475850

processor/src/test/java/com/linecorp/decaton/processor/processors/BatchingProcessorTest.java
ryamagishi: I fixed it. a882452

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
ryamagishi: I fixed it. 43f96d8

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
ryamagishi: I fixed it. 3475850

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
ryamagishi: The next `scheduleFlush()` calling is after `processBatchingTasks(batch)` is complete in same thread, so it's ok.
Is this correct? If I have a misunderstood, please point it out.🙇

ryamagishi:
@ocadaruma @kawamuray 
Sorry for delaying the reply.
I've been trying to fix the integrationTest for a week or two, but I couldn't solve it...
I need to understand how `ProcessorTestSuite` works, but that's beyond my understanding😭  
I'll continue to make efforts after this comment, but I'd appreciate it if you could tell me why the [testBatchingProcessor_capacity](https://github.com/line/decaton/pull/139/files#diff-01619b41bb5b624f7e99b97180e1ab3e48ec377aa8629c08e32932384a87b3feR70) fails. (Even if I extend timeout, it will time out.)

I have corrected other comments I received. 
As soon as this PR is merged, I will add a documentation like task-compaction.adoc in another PR.

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
ryamagishi: I fixed it. f6da8c6

ocadaruma:
Thanks for the fix.

> why the [testBatchingProcessor_capacity](https://github.com/line/decaton/pull/139/files#diff-01619b41bb5b624f7e99b97180e1ab3e48ec377aa8629c08e32932384a87b3feR70) fails

I guess it's simply because linger-based flush is disabled on that test?
So, unless each processor instance receives exactly multiple of 100 (batch size) of tasks, few remaining tasks will be never get flushed.

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
```diff
@@ -0,0 +1,142 @@
...
+public abstract class BatchingProcessor<T> implements DecatonProcessor<T> {
+
+    private final ScheduledExecutorService executor;
+    private List<BatchingTask<T>> currentBatch = Collections.synchronizedList(new ArrayList<>());
+    private final long lingerMillis;
+    private final int capacity;
+
+    @Value
+    @Accessors(fluent = true)
+    @RequiredArgsConstructor
```
ocadaruma, line 51: `@RequiredArgsConstructor` is redundant as there's `@Value`

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * An Abstract {@link DecatonProcessor} to Batch several tasks of type {@code T} to {@code List<T>}
+ * and process them at once. e.g. when downstream-DB supports batching I/O (which often very efficient).
+ * Batch-flushing is done in time-based and size-based.
+ * @param <T> type of task to batch
+ */
+public abstract class BatchingProcessor<T> implements DecatonProcessor<T> {
+
+    private final ScheduledExecutorService executor;
+    private List<BatchingTask<T>> currentBatch = Collections.synchronizedList(new ArrayList<>());
```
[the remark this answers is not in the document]
ryamagishi: I'm sorry, I misunderstood.
I fixed it.
08ed867, 242898b

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
ryamagishi: I fixed it.
aa51f27

ryamagishi:
> I guess it's simply because linger-based flush is disabled on that test?
So, unless each processor instance receives exactly multiple of 100 (batch size) of tasks, few remaining tasks will be never get flushed.

Thank you for the advice.
I tried setting `NUM_KEYS`, `NUM_SUBSCRIPTION_INSTANCES`, `NUM_PARTITIONS` in `ProcessorTestSuite` to 1, but `testBatchingProcessor_capacity()` still failed...
I removed it for now.
ref: https://github.com/line/decaton/blob/master/testing/src/main/java/com/linecorp/decaton/testing/processor/ProcessorTestSuite.java#L102-L104
7821641

ryamagishi:
Sorry for the late reply every time.
I will try to improve 🙇

ocadaruma:
Thanks for the update!

> but testBatchingProcessor_capacity() still failed

Ah maybe what I meant was more simple.
- In `testBatchingProcessor_capacity` test, lingerMillis was set to Integer.MAX.
- It means that batch-flushing is done by only size-based.
- Let's see this by example. Let's say we set `capacity` to 2 and `lingerMillis` to Int.MAX. Then what happens we feed 3 tasks?
  * 2 tasks will be processed because it will reach the capacity
  * However, remaining 1 task will be never get flushed, so it causes consumption stuck.

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
```diff
@@ -0,0 +1,139 @@
...
+    @Accessors(fluent = true)
+    public static class BatchingTask<T> {
+        Completion completion;
+        ProcessingContext<T> context;
+        T task;
+    }
+
+    /**
+     * Instantiate {@link BatchingProcessor}.
+     * If you only need one limit, please set large enough value to another.
```
ocadaruma, line 57: As described in https://github.com/line/decaton/pull/139#issuecomment-1073375516, we found that consumption will get stuck if we specify `lingerMillis` to Int.MAX.

Hm, so I think realistically we always specify non-Int.MAX value for both parameters.
Then let's just remove this line? To avoid mis-use of these parameters.

processor/src/it/java/com/linecorp/decaton/processor/BatchingProcessorTest.java
```diff
@@ -0,0 +1,67 @@
...
+    @Rule
+    public RandomRule randomRule = new RandomRule();
+
+    @Test(timeout = 30000)
+    public void testBatchingProcessor_lingerMillis() throws Exception {
+        Random rand = randomRule.random();
+        ProcessorTestSuite
+            .builder(rule)
+            .configureProcessorsBuilder(builder -> builder.thenProcess(
+                new BatchingProcessor<TestTask>(1000, Integer.MAX_VALUE) {
```
ocadaruma, line 47: I suppose the purpose of this integration test is to verify the behavior of BatchingProcessor in overall (rather than checking if BatchingProcessor is implemented correctly for each individual cases like linger=Max, capacity=Max etc), so let's also specify non-Int.MAX value for capacity here?

processor/src/main/java/com/linecorp/decaton/processor/processors/BatchingProcessor.java
ryamagishi: Exactly, I fixed it.
0cb57f7

processor/src/it/java/com/linecorp/decaton/processor/BatchingProcessorTest.java
ryamagishi: I fixed it. (I renamed the test name too.)
0317258

ryamagishi:
Thank you for your polite explanation every time!
If `ProcessorTestSuite#numTasks = 10000` and`BatchingProcessor#capacity = 100`(and `ProcessorTestSuite#NUM_SUBSCRIPTION_INSTANCES = 1`), I thought it would succeed because the number is divisible.
However, I noticed that the number of tasks processed is actually not 10000. As you said, this doesn't work well.
I'm sorry that the pasted image is difficult to understand, my question has been solved.
スクリーンショット 2022-03-22 14 29 12

ocadaruma:
What I immediately thought is because of multiple partitions.
Since the partition count is set to 8 in `ProcessorTestSuite`, if we instantiate only 1 subscription instance, it will create 8 partition-processor. With partition.concurrency = 1 and ProcessorScope = THREAD, it will create 8 batching processor instances, and clearly the received tasks count will not be multiple of 100.

ocadaruma (approved):
LGTM.

Thanks for your great work!

@kawamuray Please check the change for your comment.

kawamuray (approved):
LGTM, thanks 👍

ryamagishi:
May I merge it?

ryamagishi:
As soon as this PR is merged, I will add a documentation like task-compaction.adoc in another PR!

Issue #37, opened 2020-04-24
Change default value of max.pending.records


kawamuray:
There was a case of batching consumer that was experiencing low throughput due to too small default max.pending.records caused frequently pause/resume loop and lowering throughput.
We set it conservatively low to avoid causing high memory pressure by default setting expecting the case where average record size is large, but maybe we can default much higher expecting in most cases record size is not so big and there's more likelihood of people get benefit of high throughput with out of the box setup.

ocadaruma:
Another reasoning:

- Currently, we set consumer's `max.poll.records` to at most `100` here
  * https://github.com/line/decaton/blob/master/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ConsumerSupplier.java#L33
- This means, with default max.pending.records `100`, Decaton could pause a partition easily when single `poll()` returns 100 records

from-unknown:
I'll try this issue.

kawamuray:
Close by #143

PR #143, opened 2021-12-09
Increase max.pending.records to 10000

from-unknown:
# Summary 

Related to #37.
Because of the default `decaton.max.pending.records` is conservatively low, Decaton's throughput is not very efficient by default value.
To begin from the conclusion, based on benchmark result, `10_000` seems a suitable property for default value.

# Detailed Observations

**CAUTION : I ran benchmark on my MacBook Pro, so the result may different from real environment.**

```
Environment: MacBook Pro (16-inch, 2019)
Processor: 2.6 GHz 6-Core Intel Core i7
Memory: 16GB
```

I attached 5 benchmark result files.

attached files:
[Benchmark_max_pending_100.txt](https://github.com/line/decaton/files/7684995/Benchmark_max_pending_100.txt)
[Benchmark_max_pending_1000.txt](https://github.com/line/decaton/files/7684996/Benchmark_max_pending_1000.txt)
[Benchmark_max_pending_10000.txt](https://github.com/line/decaton/files/7684998/Benchmark_max_pending_10000.txt)
[Benchmark_max_pending_10000_2.txt](https://github.com/line/decaton/files/7684997/Benchmark_max_pending_10000_2.txt)
[Benchmark_max_pending_100000_2.txt](https://github.com/line/decaton/files/7684999/Benchmark_max_pending_100000_2.txt)

## Benchmark 1 - 100, 1,000, 10,000

Based on these results, performance is significantly changed when increase the number of max pending records.
See results below:

Common benchmark settings:

```
--title test --tasks 20000 --warmup 10000 --runs 3
```

### decaton.max.pending.records = 100

```
--- Performance ---
Execution Time(ms): 1190.33
Throughput: 19109.72 tasks/sec
Delivery Latency(ms): mean=281 max=1152
```

### decaton.max.pending.records = 1000

```
--- Performance ---
Execution Time(ms): 575.00
Throughput: 35310.44 tasks/sec
Delivery Latency(ms): mean=269 max=501
```

### decaton.max.pending.records = 10000

```
--- Performance ---
Execution Time(ms): 498.00
Throughput: 40348.33 tasks/sec
Delivery Latency(ms): mean=234 max=436
```

### Observation

As you can see, when increasing the number of max pending records, throughput and delivery latency increased.
There is a huge difference between 100 and 1,000.
Difference between 1,000 and 10,000 is not very large but still throughput increased almost 5,000 task/sec.

## Benchmark 2 - 10,000, 100,000

I also wanted to make sure that if set max pending records to 100,000, throughput will improve or not.
So I tested and results are below.

Common benchmark settings:

```
--title test --tasks 200000 --warmup 100000 --runs 3
```

### decaton.max.pending.records = 10000

```
--- Performance ---
Execution Time(ms): 4264.33
Throughput: 47047.42 tasks/sec
Delivery Latency(ms): mean=1930 max=3747
```

### decaton.max.pending.records = 100000

```
--- Performance ---
Execution Time(ms): 4383.33
Throughput: 45798.08 tasks/sec
Delivery Latency(ms): mean=2091 max=4047
```

### Observation

From the result, increasing the number to 100,000 did not make any performance improve.
Therefore, the suitable number for default is 10,000.

### Memory consumption

Increasing the default max pending records will increase memory usage.
But if we assume one task use 500B, then max memory usage for pending records will be:

```
500B * 10,000 tasks = 5,000,000B =~ 5MB
```

and this is not a problem at all.

### Crush recovery

The other downside is that if the consumer (this case Decaton) crushes, we might have possibility to run a task twice because we might not able to track tasks status.
But since Decaton enables at-least-once property by default, duplicated processing problem must be handled by user side.
Because of that, this downside is considered to not a big problem.

# Conclusion

From these observations, I suggest that set the default max pending records to 10000.
Of course the property must be tuned by each environment, increasing this number will be more likely to improve out of box performance.

### DIFF
--- a/processor/src/it/java/com/linecorp/decaton/processor/CoreFunctionalityTest.java
+++ b/processor/src/it/java/com/linecorp/decaton/processor/CoreFunctionalityTest.java
@@ -181,7 +181,8 @@ public class CoreFunctionalityTest {
                     Thread.sleep(rand.nextInt(10));
                 }))
                 .propertySupplier(StaticPropertySupplier.of(
-                        Property.ofStatic(ProcessorProperties.CONFIG_PARTITION_CONCURRENCY, 1)
+                        Property.ofStatic(ProcessorProperties.CONFIG_PARTITION_CONCURRENCY, 1),
+                        Property.ofStatic(ProcessorProperties.CONFIG_MAX_PENDING_RECORDS, 100)
                 ))
                 .customSemantics(noDuplicates)
                 .build()
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorProperties.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorProperties.java
@@ -82,7 +82,7 @@ public class ProcessorProperties extends AbstractDecatonProperties {
      * Reloadable: no
      */
     public static final PropertyDefinition<Integer> CONFIG_MAX_PENDING_RECORDS =
-            PropertyDefinition.define("decaton.max.pending.records", Integer.class, 100,
+            PropertyDefinition.define("decaton.max.pending.records", Integer.class, 10_000,
                                       v -> v instanceof Integer && (Integer) v > 0);
     /**
      * Interval in milliseconds to put in between offset commits.


kawamuray:
Thanks for the PR and thorough investigation for the change!

Your rationale based on actual benchmark result seems totally reasonable and I agree with the conclusion.

Few points leaving some comments just for note: 

>  But if we assume one task use 500B, then max memory usage for pending records will be:
> ...and this is not a problem at all.

Really depends on task size, and while a typical task serialized into hundreds of bytes, it is also common to see over-1kb task depending on its content.

Also, that 5MB isn't a problem indeed only as it is, but since single decaton subscription typically handles multiple partitions, it needs to be multiplied by the number of partitions assigned, so it could get multiple times bigger.

but yeah, even in case it should be around tens of megabytes which should still be fine in modern server instances that has tens of gigabytes memory.

so overall, LGTM 👍 

can you please promote this PR from draft PR? then I can merge it.

kawamuray:
> can you please promote this PR from draft PR? then I can merge it.

I could do that by myself :p

from-unknown:
Sorry for taking long time to fix broken test!
After discussion, I decided to set CONFIG_MAX_PENDING_RECORDS to 100 when running the failing test.

Here is the reason why the test failed and how to fix it.

The reason why the test broken:
- CONFIG_MAX_PENDING_RECORDS increased, re-balancing time also increased
- it causes timeout while re-balancing and tasks run multi-times
- some tasks ran multi-times and the test failed because this test is checking no duplicates

How to fix test:
- increase CONFIG_GROUP_REBALANCE_TIMEOUT_MS
  - when I tested, I needed to set 5000L to pass test, and this is too long for production use
- set CONFIG_MAX_PENDING_RECORDS to low value
  - if CONFIG_MAX_PENDING_RECORDS is low enough, then re-balancing will not timeout

After talking to @ocadaruma san, second way seems better because CONFIG_GROUP_REBALANCE_TIMEOUT_MS must be adjusted by user and duplicated tasks must be handled by user.
So I fixed test by set CONFIG_MAX_PENDING_RECORDS to 100 only for this test.

kawamuray (approved):

PR #147, opened 2021-12-20
Upgrade shadow which has newer log4j2 dependency

ocadaruma:
## Motivation
- Upgrade shadow plugin which has newer log4j2 dependency that is patched for recent RCE vulnerability
    * https://github.com/johnrengelman/shadow/releases/tag/7.1.1
- Note that this does **NOT** mean Decaton has log4j2 as the dependency


kawamuray:
Nice catch. Thanks!

PR #76, opened 2020-12-09
Add a workaround to build on arm64 mac

ocadaruma:
## Summary
- Decaton doesn't compile on arm64 mac
  * because protoc binary hasn't been published yet

```
% ./gradlew :processor:check
> Task :common:compileJava UP-TO-DATE
> Task :common:processResources NO-SOURCE
> Task :common:classes UP-TO-DATE
> Task :common:jar UP-TO-DATE
> Task :protocol:extractIncludeProto UP-TO-DATE
> Task :protocol:extractProto UP-TO-DATE
> Task :protocol:generateProto FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':protocol:generateProto'.
> Could not resolve all files for configuration ':protocol:protobufToolsLocator_protoc'.
   > Could not find protoc-osx-aarch_64.exe (com.google.protobuf:protoc:3.3.0).
     Searched in the following locations:
         https://repo.maven.apache.org/maven2/com/google/protobuf/protoc/3.3.0/protoc-3.3.0-osx-aarch_64.exe
```

PR #148, opened 2022-03-09
After protobuf plugin upgrade, workaround is no longer necessary

ocadaruma:
## Summary

- We added a workaround for Apple M1 in this PR https://github.com/line/decaton/pull/76
- However, I found that it's no longer necessary as already addressed in protobuf-plugin side
  * https://github.com/protocolbuffers/protobuf/pull/8557
- Forgot to remove this workaround after the plugin dep was upgraded


kawamuray:
LGTM, thanks.

PR #150, opened 2022-04-10
Add task batching doc

ryamagishi:
# Motivation
- To add a documentation for `BathingProcessor` like [task-compaction.adoc](https://github.com/line/decaton/blob/master/docs/task-compaction.adoc).

Related Issue: https://github.com/line/decaton/issues/128
Related PR: https://github.com/line/decaton/pull/139

### DIFF
--- a/docs/example/build.gradle
+++ b/docs/example/build.gradle
@@ -7,7 +7,7 @@ plugins {
 
 ext {
     DECATON_VERSION = getProperty("version") + (getProperty("snapshot").toBoolean() ? "-SNAPSHOT" : "")
-    PROTOBUF_VERSION = "3.3.0"
+    PROTOBUF_VERSION = "3.22.3"
     CENTRALDOGMA_VERSION = "0.52.5"
     KAFKA_VERSION = "2.4.0"
 }
--- a/docs/task-batching.adoc
+++ b/docs/task-batching.adoc
@@ -1,0 +1,97 @@
+= Task Batching
+:base_version: 4.0.0
+:modules: processor
+
+== Introduction
+Task Batching is a feature to batch several tasks of type `T` to `List<T>` and process them at once.
+
+Decaton provides a *time-based* way and *size-based* way to batch multiple tasks.
+The *time-based* means that tasks in past linger time are processed on every linger time.
+The *size-based* means that tasks in past before reaching capacity are processed every time tasks’size reaches capacity.
+
+== When to use (Example)
+When downstream-DB supports batching I/O (which often very efficient)
+
+== Usage
+To use `Task Batching`, you only need to instantiate `BatchingProcessor`.
+
+=== BatchingProcessor
+`BatchingProcessor` is an abstract `DecatonProcessor` to batch several tasks of type `T` to `List<T>`. To instantiate this class, pass two arguments to constructor and implement one abstract method:
+
+|===
+|parameter |Description
+
+|`lingerMillis`
+|Time limit for this processor. On every lingerMillis milliseconds, tasks in past lingerMillis milliseconds are pushed to `BatchingProcessor#processBatchingTasks(List)`.
+
+|`capacity`
+|Capacity size limit for this processor. Every time tasks’size reaches capacity, tasks in past before reaching capacity are pushed to `BatchingProcessor#processBatchingTasks(List)`.
+|===
+
+|===
+|abstract method |Description
+
+|`void processBatchingTasks(List<BatchingTask<T>> batchingTasks)`
+|When the size or time reach each limit, this method is called with stored `batchingTasks`. +
+After complete processing batch of tasks, *MUST* call `BatchingTask#completion` 's `DeferredCompletion#complete()` or `BatchingTask#context` 's `ProcessingContext#retry()` method for each `BatchingTask`. +
+The above methods is not called automatically even when an error occurs in this method, so design it so that they are called finally by yourself. Otherwise, consumption will stick. +
+`BatchingProcessor` realizes its function by using `ProcessingContext#deferCompletion()`. Reading `ProcessingContext#deferCompletion()` 's description will help you. +
+This method runs in different thread from the `process(ProcessingContext<T>, T)` thread.
+|===
+
+Before getting started with the example, let's create a `HelloTask` domain object which can be used to simulate the scenario we described:
+
+[source,java]
+.HelloTask.java
+----
+public class HelloTask {
+    private String name;
+    private int age;
+    private String text;
+
+    // ... (Getter and Setter)
+}
+----
+Instantiate `BatchingProcessor`, as shown in the following example `BathingProcessor`:
+
+[source,java]
+.TaskBatchingMain.java
+----
+public class TaskBatchingMain {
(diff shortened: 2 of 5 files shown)


docs/task-batching.adoc
```diff
@@ -0,0 +1,97 @@
...
+
+Decaton provides a *time-based* way and *size-based* way to batch multiple tasks.
+The *time-based* means that tasks in past linger time are processed on every linger time.
+The *size-based* means that tasks in past before reaching capacity are processed every time tasks’size reaches capacity.
+
+== When to use (Example)
+When downstream-DB supports batching I/O (which often very efficient)
+
+== Usage
+To use `Task Batching`, you only need to instantiate `BatchingProcessor`.
```
ocadaruma, line 16: Practically, developers will create their own class that inherits BatchingProcessor rather direct instantiation I guess.
So `you only need to implement BatchingProcessor` sounds more appropriate?

docs/task-batching.adoc
```diff
@@ -0,0 +1,97 @@
...
+                SubscriptionBuilder.newBuilder("my-decaton-processor")
+                                   .processorsBuilder(
+                                           ProcessorsBuilder
+                                                   .consuming("my-decaton-topic", extractor)
+                                                   .thenProcess(TaskBatchingMain.createBatchingProcessor(lingerMillis, capacity),
+                                                                ProcessorScope.THREAD)
+                                   )
+    }
+
+    private static BatchingProcessor<HelloTask> createBatchingProcessor(long lingerMillis, int capacity) {
```
ocadaruma, line 75: As above comment, I guess the practical usage would be implementing a class that inherits BatchingProcessor.

Then how about fixing this example as like so?

docs/task-batching.adoc
```diff
@@ -0,0 +1,97 @@
...
+`BatchingProcessor` realizes its function by using `ProcessingContext#deferCompletion()`. Reading `ProcessingContext#deferCompletion()` 's description will help you. +
+This method runs in different thread from the `process(ProcessingContext<T>, T)` thread.
+|===
+
+Before getting started with the example, let's create a `HelloTask` domain object which can be used to simulate the scenario we described:
+
+[source,java]
+.HelloTask.java
+----
+public class HelloTask {
```
ocadaruma, line 47: Let's make this example more looks like "real", by naming this class as "InsertHelloTask" or something? :)

docs/task-batching.adoc
```diff
@@ -0,0 +1,97 @@
...
+<1> Pass `lingerMillis` and `capacity` to the constructor.
+<2> Implement `processBatchingTasks(List)`
+<3> Call `BatchingTask#completion` 's `DeferredCompletion#complete()`.
+
+== Implementation
+In this section, we will briefly explain how is Task Batching implemented.
+All the magic happened in `BatchingProcessor` when a task comes to this processor the following things happen:
+
+1. The task will be put into an in-memory window.
+2. When the size or time reach each limit, `processBatchingTasks(List)` is called with stored `batchingTasks`.
```
ocadaruma, line 97: I think what we should include in "Implementation" section is something that developers should notice which comes from its implementation detail (rather than its "API"), rather than just a "how this feature is implemented".

e.g. In rate-limiting doc, it's mentioned that it adopts token-bucket algorithm which allows some "bursting" against configured rate-limit, so developers can notice that it's not suitable when they have to limit the rate strictly. (https://github.com/line/decaton/blob/master/docs/rate-limiting.adoc#implementation)

Then, we don't need to have Implementation section for BatchingProcessor I think? Or do you have some ideas you want to mention in this section?

ocadaruma (commented):
Thanks for the follow-up!

Left some feedbacks.
Also would you mind updating example project? (https://github.com/line/decaton/tree/master/docs/example/src/main/java/example)

ocadaruma:
@ryamagishi Hi, long time no see. Do you still have a plan to complete this PR?

ryamagishi:
Hi @ocadaruma, I apologize for the delay and for forgetting about this PR. If you would permit, I would like to complete it by the end of this month. I need some time to refresh my memory on the work, so I appreciate your understanding.

ocadaruma:
>  If you would permit, I would like to complete it by the end of this month

That's totally fine. Thank you!

docs/task-batching.adoc
ryamagishi: The description and examples have been updated to use `InsertHelloTaskBatchingProcessor`, which inherits `BatchingProcessor<HelloTask>`.

docs/task-batching.adoc
ryamagishi: I did the same. https://github.com/line/decaton/pull/150#discussion_r1693909020

docs/task-batching.adoc
ryamagishi: Sorry, maybe I don't understand this comment properly.

I created `InsertHelloTaskBatchingProcessor` and wrote the insertion points in the commet, but please let me know if there are any problems.
https://github.com/line/decaton/pull/150/files#diff-6cfdc5f934448411f9f9ff81c4cf114233a240ab01765dec212c7eac322f9961R70

docs/task-batching.adoc
ryamagishi: Thanks for the detailed explanation.
Removed the "Implementation" section.

ryamagishi:
@ocadaruma 
I apologize for the late response.
I have corrected the parts you commented on.
If there is anything that needs to be corrected, I will deal with it right away so that it can be merged.
Please check it when you have time.

docs/example/build.gradle
```diff
@@ -7,7 +7,7 @@ plugins {
 
 ext {
     DECATON_VERSION = getProperty("version") + (getProperty("snapshot").toBoolean() ? "-SNAPSHOT" : "")
-    PROTOBUF_VERSION = "3.3.0"
+    PROTOBUF_VERSION = "3.22.3"
```
ryamagishi, line 10: The example build failed, so I upgraded it.

docs/task-batching.adoc
ocadaruma, line 1: Could you add `Caution:` somewhere to mention below?:
- batch-flush is done in `BatchingProcessor`'s scheduled executor thread 
- Which means, the parallelism of flushing has to be controlled by ProcessorScope, not only by `decaton.partition.concurrency` config. i.e.:
  * parallelize flushing per partition: ProcessorScope.PARTITION
  * parallelize flushing per processor thread: ProcessorScope.THREAD

Some users pointed out that this behavior might be confusing, so it's worth to mention.

ocadaruma (commented):
Thanks for update, current content LGTM!

I left one more request. Please take a look.

docs/task-batching.adoc
ryamagishi: Thanks! I've added it.
https://github.com/line/decaton/pull/150/commits/7b0112a339d51d68e925a8b772058a481bb8476f

ryamagishi:
@ocadaruma 
I have responded to your comments.
Please take a look.

ocadaruma (approved):
Thank you! LGTM

PR #152, opened 2022-05-27
Test on Java 17

ocadaruma:
## Summary
- Java 17 has already been released, so the usage is expected to be spread in the future
- Add test-target JDK to confirm it works on JDK 17


kawamuray:
LGTM

PR #154, opened 2022-05-27
Upgrade kafka-clients dependency to 3.2.0

ocadaruma:
## Summary
- Upgrade kafka-clients dependency to the latest version
- But should still work with kafka 2.x
- After https://github.com/apache/kafka/pull/10203, several transitive dependencies that decaton currently depend on are moved from compile-classpath to runtime-classpath (`implementation` gradle config), we need to explicitly add dependencies for them
- Concern:
  * Should this change marked as "breaking change" and bump major version?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,7 +35,7 @@ subprojects {
     ext {
         slf4jVersion = "1.7.32"
         protobufVersion = "3.3.0"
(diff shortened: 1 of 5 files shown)


kawamuray:
LGTM, thanks

PR #168, opened 2022-07-28
Increment error counter for synchronous failure

cormoran:
After [this change](https://github.com/line/decaton/pull/99/files#diff-4df160cb6c95ec9c8815d2af1dda8a9ebe841c74c4e094bd3ef8e75074dd0af5L121) in  https://github.com/line/decaton/pull/99, decaton stopped incrementing `taskMetrics.tasksError` for synchronous failure.
This PR wants to fix `tasksError` to work again.

I applied quick patch as first PR, but implementing `completeExceptionally(Throwable)` in [DeferredCompletion](https://github.com/line/decaton/blob/d9543b907cb6af8090b4159eb7cd8fddb4cabac1/processor/src/main/java/com/linecorp/decaton/processor/DeferredCompletion.java) and complete it with exception sounds the right way.
How do you think?

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/CompletionImpl.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/CompletionImpl.java
@@ -41,6 +41,12 @@ public class CompletionImpl implements Completion {
         return comp;
     }
 
+    static CompletionImpl failedCompletion(Throwable ex) {
+        CompletionImpl comp = new CompletionImpl();
+        comp.completeExceptionally(ex);
+        return comp;
+    }
+
     public CompletionImpl() {
         future = new CompletableFuture<>();
     }
@@ -91,5 +97,10 @@ public class CompletionImpl implements Completion {
     @Override
     public void complete() {
         future.complete(null);
+    }
+
+    // package private for now, since decaton ignores exception in almost every place.
+    void completeExceptionally(Throwable ex) {
+        future.completeExceptionally(ex);
     }
 }
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/CompletionImpl.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/CompletionImpl.java
@@ -43,7 +43,7 @@ public class CompletionImpl implements Completion {
 
     static CompletionImpl failedCompletion(Throwable ex) {
         CompletionImpl comp = new CompletionImpl();
-        comp.completeExceptionally(ex);
+        comp.future.completeExceptionally(ex);
         return comp;
     }
 
@@ -97,10 +97,5 @@ public class CompletionImpl implements Completion {
     @Override
     public void complete() {
         future.complete(null);
-    }
-
-    // package private for now, since decaton ignores exception in almost every place.
-    void completeExceptionally(Throwable ex) {
-        future.completeExceptionally(ex);
     }
 }
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ProcessPipeline.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ProcessPipeline.java
@@ -128,13 +128,9 @@ public class ProcessPipeline<T> implements AutoCloseable {
         } catch (Exception e) {
             if (e instanceof InterruptedException) {
                 Thread.currentThread().interrupt();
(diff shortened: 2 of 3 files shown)


ocadaruma (commented):
Ugh nice catching. Thanks for the patch!

> but implementing completeExceptionally(Throwable) in DeferredCompletion and complete it with exception sounds the right way.

Yeah, agree.
However `DeferredCompletion` is a  public interface and we don't except it could be completed exceptionally by users.
So the right place to add `completeExceptionally` may be inside `CompletionImpl`.

cormoran:
Implemented `completeExceptionally` in `CompletionImpl`. I made it package private to keep it internal use only.

BTW, I think propagate and logging un-caught exception thrown by async processor is nicer, since it indicates users are forgetting exception handling. Anyway, it's not main interest in this fix.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/CompletionImpl.java
```diff
@@ -41,6 +41,12 @@ public static CompletionImpl completedCompletion() {
         return comp;
     }
 
+    static CompletionImpl failedCompletion(Throwable ex) {
+        CompletionImpl comp = new CompletionImpl();
+        comp.completeExceptionally(ex);
```
ocadaruma, line 46: Hm, since `completeExceptionally` is not expected to be used from outside of the class, let's just completing the underlying CompletableFuture directly?

i.e. `comp.future.completeExceptionally(ex)`

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ProcessPipeline.java
```diff
@@ -152,6 +150,7 @@ Completion process(TaskRequest request, DecatonTask<T> task) throws InterruptedE
             processMetrics.tasksCompleteDuration.record(completeDuration);
 
             if (e != null) {
+                log.error("Uncaught exception thrown by processor {} for task {}", scope, request, e);
```
ocadaruma, line 153: Just a question: The reason to move logging here is?

ocadaruma (commented):
Thanks for updating. Left few comments.

Besides the comments, would you mind adding a test to check uncaught synchronous exception is counted as error-metric correctly?

> I think propagate and logging un-caught exception thrown by async processor is nicer

I guess this is not possible from Decaton-perspective because async-processing can be executed on arbitrary user-side threads.
Using deferred completion timeout-callback would help to identify the uncaught exception handling in async-processing: https://github.com/line/decaton/blob/master/processor/src/main/java/com/linecorp/decaton/processor/ProcessingContext.java#L119

cormoran:
Sorry for long suspending. Added integration test like existing metrics test, since metrics are globally defined and it's hard to test with smaller scope.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ProcessPipeline.java
cormoran: It is to gather exception processing code in one place & catch exceptionally completed future just in case.

>> I think propagate and logging un-caught exception thrown by async processor is nicer

> I guess this is not possible from Decaton-perspective because async-processing can be executed on arbitrary user-side threads.
Using deferred completion timeout-callback would help to identify the uncaught exception handling in async-processing: https://github.com/line/decaton/blob/master/processor/src/main/java/com/linecorp/decaton/processor/ProcessingContext.java#L119

Ah, I meant "logging un-caught exception" here as exceptionally finished future like below.

```
context.deferCompletion().completeWith(CompletableFuture.failedFuture(new RuntimeException("fail")));
```

Current default `completeWith` implementation [ignores exceptionally finished future](https://github.com/line/decaton/blob/d9543b907cb6af8090b4159eb7cd8fddb4cabac1/processor/src/main/java/com/linecorp/decaton/processor/DeferredCompletion.java#L36), but I feel it's kindful to log such error in the future. (Then, putting `log.error` here sounds more natural.)

ocadaruma: Thanks, I understood.

ocadaruma (approved):
Thank you!
LGTM

ocadaruma:
Has been published in 5.0.1 https://github.com/line/decaton/releases/tag/v5.0.1

PR #169, opened 2022-09-08
Use centraldogma 0.57

mauhiz:
Contents
- Bump dependencies for CD, Jackson, SLF4J, Logback
- Fix deprecation warnings related to CD (which recently introduced CentralDogmaRepository)

After this PR, decaton will require CD v0.55+ ( https://github.com/line/centraldogma/releases )

PR #171, opened 2022-09-21
Downgrade SLF4J / logback for compatibility

mauhiz:
Follows #169


ocadaruma (approved):
Thanks for the follow up.

decaton has a spring-boot integration so it makes sense to care about transitive dependency's compatibility with boot.
LGTM.

PR #174, opened 2022-11-10
Support cooperative rebalancing fixes #98,#118

### LINKED ISSUE #98: Support cooperative rebalancing protocol

ocadaruma:
## Summary
- Kafka 2.4 added [cooperative rebalancing](https://cwiki.apache.org/confluence/display/KAFKA/KIP-429%3A+Kafka+Consumer+Incremental+Rebalance+Protocol), which aims to minimize well-known "stop-the-world" effect upon consumer rebalance
  * Essentially, what cooperative rebalancing achieves is to revoke/assign only necessary partitions on rebalance and allows to fetch on still-owned partitions even during rebalance.
  * We can opt-in cooperative rebalancing protocol by specifying `ConsumerConfig.PARTITION_ASSIGNMENT_STRATEGY_CONFIG` property
- However, currently, Decaton doesn't work with cooperative rebalancing for some reasons. This PR gonna fix them

## Problems of current Decaton implementation to support cooperative rebalancing
- Deprecated `Consumer#poll(long)` is used
  * It's already deprecated and should be replaced with `Consumer#poll(Duration)`
  * There are several behavior changes in `Consumer#poll(Duration)` though, ~from cooperative rebalancing-support perspective~, the most notable one is `poll(Duration)` doesn't block for entire rebalancing procedure unlike `poll(long)`
      - i.e. When rebalance is initiated, after `onPartitionRevoked` is called, `poll()` may return without waiting `onPartitionAssigned` invocation in same poll. (It will be eventually invoked in later `poll()` though)
  * Currently, Decaton is relying on the behavior that `onPartitionRevoked` and `onPartitionAssigned` are executed in same `poll()`
      - Once `onPartitionRevoked` returns, revoked partitions are no longer owned by KafkaConsumer so any `pause/resume/offset commit` may cause RuntimeException and kills subscription thread
      - However, since Decaton postpones destroying `PartitionContext`s to at `onPartitionAssigned` (this was to minimize context-recreation as much as possible. In short, Decaton itself did similar thing to cooperative-rebalancing from eager-rebalancing era)
      - Hence, we need some mechanism to remember which partitions are already revoked, until we actually destroy contexts in subsequent `onPartitionAssigned`
- Decaton relies on the eager-rebalancing's behavior that "all assigned partitions are revoked in `onPartitionsRevoked` => new assignment is added  in `onPartitionsAssigned`".
  * With cooperative-rebalancing, `onPartitionsAssigned` will be called only on newly-added partitions

## Summary of changes
- Store `revoking` status in `PartitionContext`, which is used to prevent calling `pause/resume/offset commit` for revoked partitions during rebalance
- Fix `ConsumeManager` to work even if subset of partitions (i.e. not full assignment) are passed in `ConsumerRebalanceListener`'s callbacks
- `SubscriptionStateListener` state transition may change
  * Because `poll()` may return after `onPartitionsRevoked` before `onPartitionsAssigned`, the transition `REBALANCING` -> `SHUTTING_DOWN` is now possible.
  * This should be considered as **breaking change**

## Appendix
- Since the behavior change that "`onPartitionsRevoked` and `onPartitionsAssigned` may be called in different `poll()` introduces certain complexity, I wrote small `TLA+` spec to get a confidence about the design
  * https://gist.github.com/ocadaruma/226c2f4e65cb97aba3610341903955b0

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,7 +35,7 @@ subprojects {
     ext {
         // Freezing this to 1.X until https://github.com/spring-projects/spring-boot/issues/12649 is resolved
         slf4jVersion = "1.7.36"
-        protobufVersion = "3.3.0"
+        protobufVersion = "3.21.7"
         kafkaVersion = "3.2.0"
         micrometerVersion = "1.7.5"
         lombokVersion = "1.18.22"
--- a/build.gradle
+++ b/build.gradle
@@ -35,7 +35,7 @@ subprojects {
     ext {
         // Freezing this to 1.X until https://github.com/spring-projects/spring-boot/issues/12649 is resolved
         slf4jVersion = "1.7.36"
-        protobufVersion = "3.21.7"
+        protobufVersion = "3.3.0"
         kafkaVersion = "3.2.0"
         micrometerVersion = "1.7.5"
         lombokVersion = "1.18.22"
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
@@ -32,7 +32,6 @@ import org.apache.kafka.clients.consumer.CommitFailedException;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.common.TopicPartition;
-import org.apache.kafka.common.errors.RebalanceInProgressException;
 import org.apache.kafka.common.errors.TimeoutException;
 
 import com.linecorp.decaton.processor.metrics.Metrics;
@@ -266,13 +265,6 @@ public class ProcessorSubscription extends Thread implements AsyncShutdownable {
             contexts.updateHighWatermarks();
             try {
                 commitManager.commitSync();
-            } catch (RebalanceInProgressException e) {
-                // At this point, Decaton already stopped processing though, from KafkaConsumer perspective,
-                // it's not closed yet (i.e. rebalance due to member-leave is not initiated), so another rebalance might be
-                // initiated (likely by other consumer instances), and then, this commitSync may fail due to
-                // RebalanceInProgressException.
-                // So, since it's not a fatal situation and kind of expected failure, we just log at WARN level.
-                log.warn("Offset commit failed because rebalance is ongoing when closing consumer", e);
             } catch (RuntimeException e) {
                 log.error("Offset commit failed before closing consumer", e);
             }
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
@@ -32,6 +32,7 @@ import org.apache.kafka.clients.consumer.CommitFailedException;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.errors.RebalanceInProgressException;
 import org.apache.kafka.common.errors.TimeoutException;
 
 import com.linecorp.decaton.processor.metrics.Metrics;
@@ -74,7 +75,7 @@ public class ProcessorSubscription extends Thread implements AsyncShutdownable {
 
     class Handler implements ConsumerHandler {
         @Override
-        public void prepareForRebalance() {
+        public void prepareForRebalance(Collection<TopicPartition> revokingPartitions) {
             updateState(SubscriptionStateListener.State.REBALANCING);
 
             waitForRemainingTasksCompletion(rebalanceTimeoutMillis.value());
@@ -83,6 +84,8 @@ public class ProcessorSubscription extends Thread implements AsyncShutdownable {
             } catch (CommitFailedException | TimeoutException e) {
                 log.warn("Offset commit failed at group rebalance", e);
             }
+
+            contexts.markRevoking(revokingPartitions);
         }
 
         @Override
@@ -263,6 +266,13 @@ public class ProcessorSubscription extends Thread implements AsyncShutdownable {
             contexts.updateHighWatermarks();
             try {
                 commitManager.commitSync();
+            } catch (RebalanceInProgressException e) {
+                // At this point, Decaton already stopped processing though, from KafkaConsumer perspective,
+                // it's not closed yet (i.e. rebalance due to member-leave is not initiated), so another rebalance might be
+                // initiated (likely by other consumer instances), and then, this commitSync may fail due to
+                // RebalanceInProgressException.
+                // So, since it's not a fatal situation and kind of expected failure, we just log at WARN level.
+                log.warn("Offset commit failed because rebalance is ongoing when closing consumer", e);
             } catch (RuntimeException e) {
                 log.error("Offset commit failed before closing consumer", e);
             }
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/AssignmentManager.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/AssignmentManager.java
@@ -52,9 +52,22 @@ public class AssignmentManager {
     interface AssignmentStore {
         /**
          * Return set of {@link TopicPartition}s that are currently assigned.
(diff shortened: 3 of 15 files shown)


build.gradle
```diff
@@ -35,7 +35,7 @@ subprojects {
     ext {
         // Freezing this to 1.X until https://github.com/spring-projects/spring-boot/issues/12649 is resolved
         slf4jVersion = "1.7.36"
-        protobufVersion = "3.3.0"
+        protobufVersion = "3.21.7"
```
mauhiz, line 38: I guess it's better to not mix dependency changes and functional changes, if avoidable?

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/AssignmentManager.java
```diff
@@ -52,10 +52,23 @@ static class AssignmentConfig {
...
 
+        /**
+         * Mark the partitions as revoking
+         * @param partitions target partition to mark
+         */
+        void markRevoking(Collection<TopicPartition> partitions);
+
+        /**
+         * Set the partitions as not revoking
+         * @param partitions target partition to unmark
```
mauhiz, line 68: partition -> partitions ?
(same above)

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java
```diff
@@ -139,11 +139,13 @@ public void destroyAllProcessors() {
...
     public Map<TopicPartition, OffsetAndMetadata> commitReadyOffsets() {
         Map<TopicPartition, OffsetAndMetadata> offsets = new HashMap<>();
-        for (PartitionContext context : contexts.values()) {
-            context.offsetWaitingCommit().ifPresent(
-                    offset -> offsets.put(context.topicPartition(),
-                                          new OffsetAndMetadata(offset + 1, null)));
-        }
+        contexts.values()
+                .stream()
+                .filter(context -> !context.revoking()).forEach(context -> {
```
mauhiz, line 144: (just to confirm) there's no issue with threading, right? same thread will update this `revoking ` flag and call `commitReadyOffsets`

mauhiz (commented):
Minimum supported version of Kafka-client should maybe be documented?

build.gradle
ocadaruma: This was to build Decaton on Apple silicon (that I'm using) on local but yeah, not necessary to include in this PR. will separate the PR

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java
ocadaruma: Yeah, all interactions are only by subscription thread so no problem.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ConsumeManager.java
```diff
@@ -129,18 +130,18 @@ public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
...
-                handler.prepareForRebalance();
+                handler.prepareForRebalance(partitions);
                 pausedPartitions.addAll(consumer.paused());
             }
 
             @Override
-            public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
-                handler.updateAssignment(partitions);
+            public void onPartitionsAssigned(Collection<TopicPartition> ignored) {
+                handler.updateAssignment(consumer.assignment());
```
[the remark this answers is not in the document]
ocadaruma: I expect so.

Also confirmed on consumer code that the only place invoking `onPartitionsAssigned` is here, and assignment is updated before invoking the callback
https://github.com/apache/kafka/blob/3.2.0/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerCoordinator.java#L457-L460

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -263,6 +266,13 @@ private void consumeLoop() {
             contexts.updateHighWatermarks();
             try {
                 commitManager.commitSync();
+            } catch (RebalanceInProgressException e) {
+                // At this point, Decaton already stopped processing though, from KafkaConsumer perspective,
+                // it's not closed yet (i.e. rebalance due to member-leave is not initiated), so another rebalance might be
+                // initiated (likely by other consumer instances), and then, this commitSync may fail due to
+                // RebalanceInProgressException.
+                // So, since it's not a fatal situation and kind of expected failure, we just log at WARN level.
```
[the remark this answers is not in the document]
ocadaruma: > "failing to commit at the end of termination" itself is an unexpected situation for users

Maybe?
However, I found that with cooperative rebalancing, "shutdown (offset commit)-during-rebalance" happens easily than eager. (I can see RebalanceInProgressException every time I run `testCooperativeRebalancing` integration test)

I'll check further and decide what would be the best here.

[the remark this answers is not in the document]
ocadaruma: > I can see RebalanceInProgressException every time I run testCooperativeRebalancing integration test

Figured out the cause.
Integration test does:
1. start 3 subscriptions one by one, with awaiting each subscription transitions to `RUNNING`
2. Start rolling restart of subscriptions. This includes stopping a subscription and then start it.

The problem is, with cooperative rebalancing, "state transitioned to `RUNNING`" is no longer a sign of rebalance stabilized.
With cooperative rebalancing, typical rebalance by new-member addition causes 2 rounds of consumer rebalance.
One for revoking the partition from existing member, one for assign them to newly added member. This is to ensure single partition never be assigned to multiple instances.

So, if step 2 above (rolling-restart phase) starts before finishing second round of member-addition-rebalance, when rolling restarting, Decaton may fail to commit-sync on shutdown due to `RebalanceInProgressException`. That's what happened in `testCooperativeRebalancing`.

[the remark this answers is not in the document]
ocadaruma: So, unless this (https://issues.apache.org/jira/browse/KAFKA-9382) is addressed, users may see more offset commit failure on application rolling restart due to this in cooperative rebalancing.

Anyways, I agree with keeping it `ERROR` because it means duplicate-processing will likely occur, which users would not expect for usual rolling restart.

kawamuray (approved):
LGTM. @mauhiz wanna continue reviewing?

ocadaruma:
> Minimum supported version of Kafka-client should maybe be documented?

Sorry, overlooked.

After this PR gets merged, Decaton will require kafka-clients at least `2.0.0` because `poll(Duration)` is used but I think it's quite long ago so not necessary to be explicitly documented?

mauhiz:
Thanks, I'm good. Thanks for the nice feature!

Regarding the minimum version of Kafka client, I think it's better documented than not!

ocadaruma:
Okay then, I'll followup in separate PR about minimum kafka-clients ver.

PR #175, opened 2022-11-23
Upgrade protobuf to support apple silicon

ocadaruma:
- My bad, I removed workaround for apple silicon build in https://github.com/line/decaton/pull/148 but forgot to upgrade the plugin, so now still failing on apple silicon


kawamuray:
LGTM

PR #176, opened 2022-12-09
Ensure timeout is visible from subscription thread

ocadaruma:
- `OffsetState#timeoutAt` is set by processor threads and reaper thread, while it's read by subscription thread when executing `OOOCC#updateHighWatermark`.
- We should make timeoutAt volatile to ensure latest timeout value is visible from subscription thread


kawamuray (approved):
nice catch, thx

Issue #167, opened 2022-07-26
Isolate bursting key to prevent innocent keys are affected


ocadaruma:
- In Kafka, it's common to use some entity's identifier (e.g. user-id) to ensure task processing ordering per entity.
- At the same time, this tends to cause "hot-spot" key when there's an abusive task generation for single entity
- When it happens, Decaton's processing thread could be occupied by such bursting key and could affect other innocent keys processing
- It's nice to have a built-in feature to detect hot-spots and isolate workload to protect innocent keys

(Actually, this is a missing long-awaited Decaton feature that necessity exists from early days...)

PR #177, opened 2022-12-15
[API draft] Per key quota

ocadaruma:
- This is the draft for the feature fixes https://github.com/line/decaton/issues/167
- I would like to discuss the API first, then proceed to actual implementation after that
- `example/PerKeyQuotaMain.java` describes the complete usage of this feature


kawamuray:
roughly checked API design and it looks good.

ocadaruma:
Thanks. Then I'll close now for recording, and proceed as another PR

PR #178, opened 2022-12-16
Add retry count metric

mosmeh:
# Motivation

In our use cases, we try to detect repeated failures of task processing by logging when tasks are retried.
However, frequent repeated logs tend to be ignored, so we may not notice repeated processing failures for a long time.
Instead, this PR introduces a retry count metric, which can be used for alerting based on thresholds.

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
@@ -263,8 +263,8 @@ public class Metrics {
                                    .tags(availableTags.subscriptionScope())
                                    .register(registry));
 
-        public final DistributionSummary retryCount =
-                meter(() -> DistributionSummary.builder("retry.count")
+        public final DistributionSummary retryTaskRetries =
+                meter(() -> DistributionSummary.builder("retry.task.retries")
                                                .description("The number of times a task was retried")
                                                .tags(availableTags.subscriptionScope())
                                                .register(registry));
--- a/processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
@@ -30,6 +30,7 @@ import org.apache.kafka.clients.consumer.Consumer;
 import com.linecorp.decaton.processor.metrics.internal.AvailableTags;
 
 import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.DistributionSummary;
 import io.micrometer.core.instrument.Gauge;
 import io.micrometer.core.instrument.Meter;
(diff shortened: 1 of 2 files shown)


processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
```diff
@@ -261,6 +262,12 @@ public class RetryMetrics extends AbstractMetrics {
                                    .description("The number of tasks failed to enqueue in retry topic")
                                    .tags(availableTags.subscriptionScope())
                                    .register(registry));
+
+        public final DistributionSummary retryCount =
+                meter(() -> DistributionSummary.builder("retry.count")
```
ajalab, line 267: If this distribution summary is exposed as Prometheus/OpenTelemetry [Histogram](https://prometheus.io/docs/concepts/metric_types/#histogram) metric, IIUC a time series with `_count` suffix will be created. This results in having `retry_count_count`, which could be redundant?

How about naming it something like `retry.task.retries`?

> ...should have a suffix describing the unit, in plural form.
https://prometheus.io/docs/practices/naming/

mosmeh: Thank you. Updated in https://github.com/line/decaton/pull/178/commits/4e704e775401c2d6475b35ff79a2bceb97e79768

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/DecatonTaskRetryQueueingProcessor.java
```diff
@@ -71,6 +72,7 @@ public void process(ProcessingContext<byte[]> context, byte[] serializedTask)
             } else {
                 metrics.retryQueueingFailed.increment();
             }
+            metrics.retryTaskRetries.record(nextRetryCount);
```
ocadaruma, line 75: Isn't it better to record before sending produce request instead of inside callback?
Because we may want to know "retry is attempted over X times"  regardless queueing is completed or not

ocadaruma (commented):
Thanks for the patch!
The motivation sounds make sense.

Left minor question but overall looks good.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/DecatonTaskRetryQueueingProcessor.java
mosmeh: Thank you, that sounds reasonable. Updated in https://github.com/line/decaton/pull/178/commits/58e93c5b33e5cbf0ce585b723aaaa018576a2dfe

ajalab (approved):

ocadaruma (approved):

PR #180, opened 2023-02-13
Add missing url on docs

Yang-33:
https://github.com/line/decaton/pull/69 added the distributed tracing feature and its document but an index page doesn't have its url. This pull request just adds the url in the index page.


ocadaruma (approved):
Nice catching. Thanks!

PR #182, opened 2023-02-15
Provide a way of specifying a partition when producing a task 

### LINKED ISSUE #159: Provide a way of specifying a partition in producing a task
We can inject a custom `org.apache.kafka.clients.producer.Partitioner` by implementing it and passing it through `DecatonClientBuilder#producerConfig`.

However, A decaton client wraps a task set by our application in `DecatonTaskRequest`.
So, to determine a partition based on a task in `Partitioner`, we should handle a `DecatonTaskRequest`  (deserialize our defined task. after that proceed partition step) like the following.

```java
public class CustomPartitioner extends DefaultPartitioner {

    @Override
    public int partition(String topic, Object key, byte[] keyBytes, Object value, byte[] valueBytes, Cluster cluster) {
        if (value instanceof DecatonTaskRequest) {
            DecatonTaskRequest request = (DecatonTaskRequest) value;
            MyTask task = new ProtocolBuffersDeserializer<>(MyTask.parser())
                                           .deserialize(request.getSerializedTask().toByteArray());
            # determine partition
            .... 
            return result;
        }
        return super.partition(topic, key, keyBytes, value, valueBytes, cluster);
    }

```

The above seems redundant and Inefficient due to deserializing a serialized task.

Could we provide a way of specifying a partition in producing a task like the following API?

```java
public class DecatonClientImpl<T> implements DecatonClient<T> {
   ....
  public CompletableFuture<PutTaskResult> put(String key, T task, int partition) {
  ...
}

```

The partition that is passed as the argument would be set to `TaskMetadataProto` and be used in an internal `Partitioner` that determines a partition base on it.

What do you think about this feature?

ta7uw:
Motivation:

To resolve https://github.com/line/decaton/issues/159

Modifications:

- Add a new API to `DecatonClient` and `DecatonTaskProducer`

Result:

- Closes #159
- Users can produce a partition-specified task without redundant code and Inefficient deserialization

### DIFF
--- a/client/src/main/java/com/linecorp/decaton/client/DecatonClient.java
+++ b/client/src/main/java/com/linecorp/decaton/client/DecatonClient.java
@@ -19,6 +19,9 @@ package com.linecorp.decaton.client;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
 
+import org.apache.kafka.clients.producer.Partitioner;
+import org.apache.kafka.common.Cluster;
+
 import com.linecorp.decaton.common.Serializer;
 
 import lombok.Builder;
@@ -78,6 +81,19 @@ public interface DecatonClient<T> extends AutoCloseable {
      * @return a {@link CompletableFuture} which represents the result of task put.
      */
     CompletableFuture<PutTaskResult> put(String key, T task, Consumer<Throwable> errorCallback);
+
+    /**
+     * Put a task onto a specified kafka partition.
+     * Note that the specified partition here is prioritised over the result of
+     * {@link Partitioner#partition(String, Object, byte[], Object, byte[], Cluster)}.
+     *
+     * @param key the criteria to shuffle and order tasks. null can be specified if it doesn't matters.
+     * @param task an instance of task. Should never be null.
+     * @param partition the id of the partition
+     *
+     * @return a {@link CompletableFuture} which represents the result of task put.
+     */
+    CompletableFuture<PutTaskResult> put(String key, T task, int partition);
 
     /**
      * Put a task onto associated decaton queue with specifying arbitrary timestamp.
--- a/client/src/main/java/com/linecorp/decaton/client/DecatonClient.java
+++ b/client/src/main/java/com/linecorp/decaton/client/DecatonClient.java
@@ -63,9 +63,11 @@ public interface DecatonClient<T> extends AutoCloseable {
 
     /**
      * Put a task onto a specified kafka partition with specifying some fields of task metadata.
-     * @param key the criteria to shuffle and order tasks. null can be specified if it doesn't matters.
+     * @param key the criteria to shuffle and order tasks. null can be specified if it doesn't matter.
      * @param task an instance of task. Should never be null.
-     * @param partition the id of the partition
+     * @param overrideTaskMetadata taskMetaData which can be set by users and used for event publish.
+     * null can be specified if it doesn't matter.
+     * @param partition the id of the partition. null can be specified if it doesn't matter.
      *
      * @return a {@link CompletableFuture} which represents the result of task put.
      */
--- a/client/src/main/java/com/linecorp/decaton/client/DecatonClient.java
+++ b/client/src/main/java/com/linecorp/decaton/client/DecatonClient.java
@@ -62,6 +62,17 @@ public interface DecatonClient<T> extends AutoCloseable {
     CompletableFuture<PutTaskResult> put(String key, T task, TaskMetadata overrideTaskMetadata);
 
     /**
+     * Put a task onto a specified kafka partition with specifying some fields of task metadata.
+     * @param key the criteria to shuffle and order tasks. null can be specified if it doesn't matters.
+     * @param task an instance of task. Should never be null.
+     * @param partition the id of the partition
+     *
+     * @return a {@link CompletableFuture} which represents the result of task put.
+     */
+    CompletableFuture<PutTaskResult> put(String key, T task, TaskMetadata overrideTaskMetadata,
+                                         Integer partition);
+
+    /**
      * Put a task onto associated decaton queue.
      * This is just a helper method for typical use cases - put task asynchronously and observe result just for telling
      * whether the message production succeeded:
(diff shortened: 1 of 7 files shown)


client/src/main/java/com/linecorp/decaton/client/DecatonClient.java
```diff
@@ -79,6 +82,19 @@
...
+     * Note that the specified partition here is prioritised over the result of
+     * {@link Partitioner#partition(String, Object, byte[], Object, byte[], Cluster)}.
+     *
+     * @param key the criteria to shuffle and order tasks. null can be specified if it doesn't matters.
+     * @param task an instance of task. Should never be null.
+     * @param partition the id of the partition
+     *
+     * @return a {@link CompletableFuture} which represents the result of task put.
+     */
+    CompletableFuture<PutTaskResult> put(String key, T task, int partition);
```
ta7uw, line 96: I have a concern about this API.
If users use the following API with `int` as the argument for `timestamp`, I think the behavior will be different after this change.
https://github.com/line/decaton/blob/916403fac80bef5b222a661d836cb814f4aa9a53/client/src/main/java/com/linecorp/decaton/client/DecatonClient.java#L50

ocadaruma: Good point, this could be an error prone for users.

Also, if we want to specify timestamp or other task metadata with partition, we have to add other overloads and it's a mess.

So, how about adding one "universal" method instead? i.e.
```
put(String key, T task, TaskMetadata overrideTaskMetadata, Integer partition)
```
With allowing setting overrideTaskMetadata or partition to be null for users who don't need to set them explicitly.

client/src/main/java/com/linecorp/decaton/client/DecatonClient.java
```diff
@@ -79,6 +82,19 @@
      */
     CompletableFuture<PutTaskResult> put(String key, T task, Consumer<Throwable> errorCallback);
 
+    /**
+     * Put a task onto a specified kafka partition.
+     * Note that the specified partition here is prioritised over the result of
+     * {@link Partitioner#partition(String, Object, byte[], Object, byte[], Cluster)}.
```
ocadaruma, line 88: I think we do not need to mention about precedence here explicitly, because this is KafkaProducer's behavior rather than DecatonClient's interface.

client/src/main/java/com/linecorp/decaton/client/DecatonClient.java
ta7uw: It seems good

client/src/main/java/com/linecorp/decaton/client/DecatonClient.java
```diff
@@ -58,6 +61,17 @@
      */
     CompletableFuture<PutTaskResult> put(String key, T task, TaskMetadata overrideTaskMetadata);
 
+    /**
+     * Put a task onto a specified kafka partition with specifying some fields of task metadata.
+     * @param key the criteria to shuffle and order tasks. null can be specified if it doesn't matters.
+     * @param task an instance of task. Should never be null.
+     * @param partition the id of the partition
```
ocadaruma, line 68: Let's mention also when partition, overrideTaskMetadata can be null

client/src/main/java/com/linecorp/decaton/client/internal/DecatonClientImpl.java
```diff
@@ -81,6 +82,19 @@ public CompletableFuture<PutTaskResult> put(String key, T task, TaskMetadata ove
         return put(key, task, convertToTaskMetadataProto(overrideTaskMetadata));
     }
 
+    @Override
+    public CompletableFuture<PutTaskResult> put(String key, T task, TaskMetadata overrideTaskMetadata,
```
ocadaruma, line 86: This method supersedes below `put(String key, T task, TaskMetadataProto taskMetadataProto)` so let's change it to just calling this method?

client/src/main/java/com/linecorp/decaton/client/internal/DecatonTaskProducer.java
```diff
@@ -63,7 +63,15 @@ public DecatonTaskProducer(String topic, Properties producerConfig,
 
     public CompletableFuture<PutTaskResult> sendRequest(byte[] key, DecatonTaskRequest request) {
         ProducerRecord<byte[], DecatonTaskRequest> record = new ProducerRecord<>(topic, key, request);
+        return sendRequest(record);
+    }
+
+    public CompletableFuture<PutTaskResult> sendRequest(byte[] key, DecatonTaskRequest request, int partition) {
```
ocadaruma, line 69: DecatonTaskProducer is in `internal` package so not considered to be public API, so we don't need to keep backward compatibility. Let's just change original `sendRequest` to receive `Integer partition`

client/src/main/java/com/linecorp/decaton/client/internal/DecatonClientImpl.java
```diff
@@ -118,6 +127,9 @@ private TaskMetadataProto convertToTaskMetadataProto(TaskMetadata overrideTaskMe
 
         final Long timestamp = overrideTaskMetadata.getTimestamp();
         final Long scheduledTime = overrideTaskMetadata.getScheduledTime();
+        if (overrideTaskMetadata == null) {
```
ocadaruma, line 130: This null check should be above `overrideTaskMetadata.getTimestamp()` .

Otherwise it causes NPE

client/src/test/java/com/linecorp/decaton/client/internal/DecatonClientImplTest.java
```diff
@@ -155,6 +156,37 @@ public void testWithEmptyTaskMetaDataSetExternally() {
...
+        assertNull(record.timestamp());
+        assertEquals(5678L, record.value().getMetadata().getTimestampMillis());
+        assertEquals(6912L, record.value().getMetadata().getScheduledTimeMillis());
+    }
+
+    @Test
+    public void testSpecifyingPartitionWithoutMetadata() {
+        doReturn(1234L).when(timestampSupplier).get();
+
+        client.put("key", HelloTask.getDefaultInstance(), TaskMetadata.builder().build(), 4);
```
ocadaruma, line 182: Isn't this test supposed to pass `null` as the taskmetadata argument? (rather than empty task metadata)

ocadaruma (commented):
Left last 2 comments, but almost lgtm

ocadaruma (approved):
LGTM. Thanks!

PR #184, opened 2023-02-15
Provide round-robin mode in SubPartitioner for non-null key (resolve #146)

### LINKED ISSUE #146: Provide round-robin mode in SubPartitioner for non-null key
- Currently, Decaton's SubPartitioner uses murmur2 hash of the key to determine the subpartition to ensure per-key processing ordering guarantee.
- However, sometimes we want to use round-robin strategy even for non-null key. e.g.:
  * Topic is consumed by multiple consumers, and some consumers need ordering guarantee (e.g. windowed-processing) but some consumers don't (just collecting stats).
  * In latter type of consumers, using hash-partitioning doesn't make sense. (just spoils the processing throughput when key-bursting happens)
- It's nice to have a feature to enable round-robin mode even for non-null keyed topics.

palindrom615:
* Add `ProcessorProperties#CONFIG_PARTITION_BY_KEY` config
* Add `SubpartitioningStrategy strategy` property in  `SubPartitioner`class

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/DefaultSubPartitioner.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/DefaultSubPartitioner.java
@@ -37,9 +37,9 @@ public class DefaultSubPartitioner implements SubPartitioner {
     }
 
     @Override
-    public int partitionFor(byte[] key) {
+    public int subPartitionFor(byte[] key) {
         if (key == null) {
-            return roundRobinStrategy.partitionFor(key);
+            return roundRobinStrategy.subPartitionFor(key);
         } else {
             // Kafka client uses murmur2 for hashing keys to decide partition to route the record,
             // so all keys we receive in a partition processor has highly biased distribution.
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/DefaultSubPartitioner.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/DefaultSubPartitioner.java
@@ -18,6 +18,11 @@ package com.linecorp.decaton.processor.runtime;
 
 import java.nio.ByteBuffer;
 
+/**
+ * Default implementation for {@link SubPartitioner}.
+ * This implementation guarantees that tasks with same key will always be assigned to the same subpartition.
+ * If the key is null, the subpartition will be chosen in round-robin manner.
+ */
 public class DefaultSubPartitioner implements SubPartitioner {
     private final int bound;
     private final RoundRobinSubPartitioner roundRobinStrategy;
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/DefaultSubPartitioner.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/DefaultSubPartitioner.java
@@ -14,36 +14,27 @@
  * under the License.
  */
 
-package com.linecorp.decaton.processor.runtime.internal;
+package com.linecorp.decaton.processor.runtime;
 
 import java.nio.ByteBuffer;
-import java.util.concurrent.atomic.AtomicLong;
 
-class SubPartitioner {
+public class DefaultSubPartitioner implements SubPartitioner {
     private final int bound;
-    private final AtomicLong monotonicValueSupplier;
-    private final SubpartitioningStrategy strategy;
+    private final RoundRobinSubPartitioner roundRobinStrategy;
 
-    SubPartitioner(int bound) {
-        this(bound, null);
-    }
-
-    SubPartitioner(int bound, SubpartitioningStrategy strategy) {
-        if (strategy == null) {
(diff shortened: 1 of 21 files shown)


ocadaruma:
For this issue, we should decide overall strategy first I guess.

There are two options to provide "always round-robin subpartitioning".

- 1. Provide a config to switch to "always round-robin" (as this PR does)
- 2. Make subpartitioner customizable, and provide `DefaultSubPartitioner` and something like `RoundRobinSubPartitioner` built-in

For the initial motivation, option 1 may suffice but I +1 for option 2 because I can imagine some use cases of custom sub partitioner.
- e.g. If key's distribution is known in user's specific usage, users could implement custom partitioner to achieve more "well balancing among threads" while preserving per-key process ordering

@kawamuray WDYT?

palindrom615:
@ocadaruma 

* WDYT about the naming `SubpartitioningStrategy`, `DefaultSubpartitioningStrategy`, and `RoundRobinSubpartitioningStrategy`?
* how do you think `SubpartitioningStrategy` should be selected? `PartitionProcessor` constructor? `PartitionContext` -> `PartitionProcessor#setSubpartitioningStrategy`? or by reading property?

ocadaruma (commented):
> WDYT about the naming SubpartitioningStrategy, DefaultSubpartitioningStrategy, and RoundRobinSubpartitioningStrategy

I prefer the name `SubPartitioner`, which is analogous to Kafka's `Partitioner` so may be more understandable.

> how do you think SubpartitioningStrategy should be selected

What I initially thought was passing `SubpartitioningStrategy` in `SubscriptionBuilder` and propagate it to the `PartitionProcessor` through `PartitionScope (which extends SubscriptionScope)` .

However, I realized that we can't just pull around concrete `SubpartitioningStrategy` instance because it needs to be instantiated per partition, every time new partition is assigned.

So I realize we need to have a way to inject `Supplier of SubpartitioningStrategy` in the SubscriptionBuilder instead.

What I imagine roughly is:
```
@FunctionalInterface
public interface SubPartitionerSupplier {
    SubPartitioner get(int numSubPartitions);
}
```
and propagate it from `SubscriptionBuilder` to `PartitionProcessor` through `PartitionScope`.

WDYT?

palindrom615:
hmm.. I just naively passed the `SubPartitionerSupplier` through `SubscriptionBuilder` -> `ProcessorSubscription` -> `PartitionContexts` -> `PartitionScope` -> `PartitionProcessor`.

@ocadaruma could you please re-review this PR?

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java
```diff
@@ -160,13 +160,24 @@ public void receive(ConsumerRecord<byte[], byte[]> record) {
         setName(String.format("DecatonSubscriptionThread-%s", scope));
     }
 
+
+    public ProcessorSubscription(SubscriptionScope scope,
```
ocadaruma, line 164: We don't need another constructor and pulling `SubPartitionerSupplier` around multiple classes.

We can just add new field in `SubscriptionScope`.

Then, in `PartitionProcessor`'s constructor, it can be accessible through `PartitionScope` constructor parameter because `PartitionScope` extends `SubscriptionScope`

processor/src/main/java/com/linecorp/decaton/processor/runtime/DefaultSubPartitioner.java
```diff
@@ -14,27 +14,27 @@
...
  */
 
-package com.linecorp.decaton.processor.runtime.internal;
+package com.linecorp.decaton.processor.runtime;
 
 import java.nio.ByteBuffer;
-import java.util.concurrent.atomic.AtomicLong;
 
-class SubPartitioner {
+public class DefaultSubPartitioner implements SubPartitioner {
```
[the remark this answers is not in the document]
palindrom615: added!

processor/src/main/java/com/linecorp/decaton/processor/runtime/SubPartitioner.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package com.linecorp.decaton.processor.runtime;
+
+/**
+ * Decides processor unit(subpartition) for a task.
+ */
+@FunctionalInterface
+public interface SubPartitioner {
+    int partitionFor(byte[] key);
```
ocadaruma, line 24: I prefer `subPartitionFor` to avoid confusion with Kafka's `partition`

ocadaruma (commented):
LGTM except only one minor point!

processor/src/main/java/com/linecorp/decaton/processor/runtime/SubPartitioner.java
palindrom615: Fixed. thanks!

ocadaruma (approved):
LGTM Thanks!

PR #186, opened 2023-02-22
Silence WARNs during rebalancing

### LINKED ISSUE #185: Silence WARNs during rebalancing
On decaton 6, observing the following WARNs, when
- cooperating rebalancing is ON
- a node joins the consumer group

```
com.linecorp.decaton.processor.runtime.internal.CommitManager:131

Offset commit failed asynchronously

org.apache.kafka.common.errors.RebalanceInProgressException: Offset commit cannot be completed since the consumer is undergoing a rebalance for auto partition assignment. You can try completing the rebalance by calling poll() and then retry the operation.
```

I don't think the end-user needs to know about this, since `poll()` is soon going to be called again in `ProcessorSubscription.consumeLoop`

mauhiz:
Closes #185

Only WARN when commit failed due a situation that needs attention

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/CommitManager.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/CommitManager.java
@@ -21,7 +21,9 @@ import java.util.Map;
 
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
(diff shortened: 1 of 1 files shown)


ocadaruma:
Thanks for your PR.

However, as discussed in #185, we would like to keep WARN level as default since it's still a sign of duplicated processing, and.
Users can configure logger setting (e.g. logback.xml) to suppress these logs if necessary.

PR #188, opened 2023-03-01
Add metrics to check DeferredCompletion leak

### LINKED ISSUE #39: Consider adding metric that shows smallest offset that is not completed
It would be useful for precise detection and spot the causing record when `DeferredCompletion` leak occurred by user-supplied code.

ta7uw:
Motivation:

We can detect `DeferredCompletion ` leak by monitoring `latest consumed offset - last committed offset < max.pending.record`.
Therefore, we need to add these metrics.

Modifications:

- Add the following metrics
  - `decaton.offset.last.committed` 
  - `decaton.offset.latest.consumed`

Result:

- Closes #39
- Users can monitor `DeferredCompletion` leak

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContext.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContext.java
@@ -55,6 +55,7 @@ public class PartitionContext implements AutoCloseable {
     @Getter
     @Setter
     private boolean revoking;
+    private long latestConsumedOffset;
 
     public PartitionContext(PartitionScope scope, Processors<?> processors, int maxPendingRecords) {
         this.scope = scope;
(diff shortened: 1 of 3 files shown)


processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContext.java
```diff
@@ -55,6 +55,7 @@ public class PartitionContext implements AutoCloseable {
     @Getter
     @Setter
     private boolean revoking;
+    private long latestConsumedOffset;
```
kawamuray, line 58: This and `lastCommittedOffset` needs to be `volatile` because they can be written and read by different threads at arbitrary timing (decaton's thread and metric exporting server's thread)

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContext.java
```diff
@@ -134,6 +136,7 @@ public void addRequest(TaskRequest request) {
     }
 
     public OffsetState registerOffset(long offset) {
+        latestConsumedOffset = Math.max(latestConsumedOffset, offset);
```
kawamuray, line 139: I think there's no chance of the offset being registered to go backward? Even in regression case, we should update `latestConsumedOffset` as is because we're actually consuming that point now?

kawamuray (changes requested):
Thanks for your PR! left a few comments

kawamuray (approved):
LGTM, thanks!

PR #191, opened 2023-03-02
Narrow down the scope of waiting for pending tasks to per partition

### LINKED ISSUE #172: Narrow the scope of waiting pending task count upon dynamic property reload
- Currently, Decaton's dynamic property reload is implemented like below:
  - When dynamic-property change is detected, up "reloadRequested" flag: https://github.com/line/decaton/blob/v5.0.1/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java#L74
  - Wait all pending tasks are processed
  - After that, recreating partition contexts with new concurrency count

Though this "safepoint"-like mechanism was to preserve per-key process ordering guarantee even on concurrency change, at least, we can narrow down the scope of safepoint to per-partition instead of global without violating per-key ordering guarantee.

ta7uw:
Motivation:

Described in https://github.com/line/decaton/issues/172

Modifications:

- Reload context per partition when reloading is requested

Result:

- Closes #172 
- Reduces time to resume processing when dynamic property reloaded

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContext.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContext.java
@@ -56,6 +56,14 @@ public class PartitionContext implements AutoCloseable {
     @Setter
     private boolean revoking;
 
+    /**
+     * Indicates that if true, reloading is requested and not completed.
+     * This is used to perform reloading processing for each partition.
+     */
+    @Getter
+    @Setter
+    private boolean reloadState;
+
     public PartitionContext(PartitionScope scope, Processors<?> processors, int maxPendingRecords) {
         this.scope = scope;
         this.processors = processors;
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContext.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContext.java
@@ -62,7 +62,7 @@ public class PartitionContext implements AutoCloseable {
      */
     @Getter
     @Setter
-    private boolean reloadState;
+    private volatile boolean reloadRequested;
 
     public PartitionContext(PartitionScope scope, Processors<?> processors, int maxPendingRecords) {
         this.scope = scope;
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContext.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContext.java
@@ -56,6 +56,9 @@ public class PartitionContext implements AutoCloseable {
     @Setter
     private boolean revoking;
 
+    @Getter
+    private boolean reloading;
+
     public PartitionContext(PartitionScope scope, Processors<?> processors, int maxPendingRecords) {
         this.scope = scope;
         this.processors = processors;
@@ -155,6 +158,13 @@ public class PartitionContext implements AutoCloseable {
         long pausedNanos = System.nanoTime() - pausedTimeNanos;
         pausedTimeNanos = -1;
         metrics.partitionPausedTime.record(pausedNanos, TimeUnit.NANOSECONDS);
+    }
+
+    public void reloading(boolean reloading) {
+        this.reloading = reloading;
+        if (reloading) {
+            pause();
+        }
     }
 
     @Override
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContext.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContext.java
@@ -56,9 +56,6 @@ public class PartitionContext implements AutoCloseable {
     @Setter
     private boolean revoking;
 
-    @Getter
-    private boolean reloading;
-
     public PartitionContext(PartitionScope scope, Processors<?> processors, int maxPendingRecords) {
         this.scope = scope;
         this.processors = processors;
@@ -158,13 +155,6 @@ public class PartitionContext implements AutoCloseable {
         long pausedNanos = System.nanoTime() - pausedTimeNanos;
         pausedTimeNanos = -1;
         metrics.partitionPausedTime.record(pausedNanos, TimeUnit.NANOSECONDS);
-    }
-
-    public void reloading(boolean reloading) {
-        this.reloading = reloading;
-        if (reloading) {
-            pause();
-        }
     }
 
     @Override
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java
@@ -22,17 +22,12 @@ import static java.util.stream.Collectors.toMap;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Function;
-import java.util.stream.Collectors;
 
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
 import org.apache.kafka.common.TopicPartition;
@@ -289,7 +284,7 @@ public class PartitionContexts implements OffsetsStore, AssignmentStore, Partiti
                     .filter(entry -> entry.getValue().reloadRequested()
                                      && entry.getValue().pendingTasksCount() == 0)
                     .map(Entry::getKey)
-                    .collect(Collectors.toList());
+                    .collect(toList());
             if (reloadableTopicPartitions.isEmpty()) {
                 return;
             }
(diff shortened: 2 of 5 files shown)


ocadaruma (commented):
Thanks for submitting the PR.

This PR seems trying to observe pending task count per-partition.
However, it's not enough because as long as `reloadRequested` is true, Decaton pauses all processing so anyways none of partitions can make progress. (https://github.com/line/decaton/blob/c4bbbc17e933000aa48d56c6b0068ad8d7af9233/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java#L215)

So we need to manage reload-requested flag per-partition I think

ta7uw:
Thanks for your review.
I'll take it into consideration.

ta7uw:
@ocadaruma 
Could you please review it again?

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContext.java
```diff
@@ -157,6 +160,13 @@ public void resume() {
         metrics.partitionPausedTime.record(pausedNanos, TimeUnit.NANOSECONDS);
     }
 
+    public void reloading(boolean reloading) {
+        this.reloading = reloading;
+        if (reloading) {
+            pause();
```
ocadaruma, line 166: pausing here by PartitionContext itself doesn't work to pause fetching actually.

Decaton's fetch-pause works like:
- (1) ConsumeManager retrieves partition list that needs pause https://github.com/line/decaton/blob/v6.2.0/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ConsumeManager.java#L182
- (2) Call Consumer#pause to stop fetching further records. From now on, Consumer#poll doesn't return records for these partitions

So, if we pause here by PartitionContext itself, it will not be included in (1) (because it's already "paused").

Correct approach here may be holding per-partition `reloading` flag in PartitionContexts, and include them in `partitionsNeedsPause`

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java
```diff
@@ -71,8 +76,16 @@ public PartitionContexts(SubscriptionScope scope, Processors<?> processors) {
                 return;
             }
 
-            if (!reloadRequested.getAndSet(true)) {
-                logger.info("Requested reload partition.concurrency oldValue={}, newValue={}", oldVal, newVal);
+            lock.lock();
+            try {
```
ocadaruma, line 80: Currently, single `reloadRequested` flag is shared among all partitions, while reloadContext may happen in different timings.

This could cause inconsistent partition concurrency among partitions as below:
- (1) `decaton.partition.concurrency` is changed from 3 to 5
- (2) `reloadRequested` is set to true
- (3) partition-X is reloaded with 5
- (4) Before partition-Y is reloaded, `decaton.partition.concurrency` is changed from 5 to 4
- (5) Since `reloadRequested` is already true, reload for partition-X doesn't happen.
- (6) partition-Y is reloaded with 4
- As the result, partition-X has 5 threads unexpectedly

I think we need per-partition `reloadRequested` flag.

ta7uw:
@ocadaruma
I fixed to hold reload request per-partition.
Could you please review it again?

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java
```diff
@@ -250,23 +271,31 @@ public void partitionsResumed(List<TopicPartition> partitions) {
...
         if (reloadRequested.get()) {
-            if (totalPendingTasks() > 0) {
-                logger.debug("Waiting pending tasks for property reload.");
-                return;
+            lock.lock();
+            try {
+                List<TopicPartition> reloadableTopicPartitions = reloadStates.entrySet()
+                        .stream()
+                        .filter(entry -> entry.getValue()
+                                         && contexts.get(entry.getKey()).pendingTasksCount() == 0)
```
ocadaruma, line 279: This could cause NPE because currently `reloadStates` entry never removed even when partition is removed.

To fix that, we can add a code to remove `reloadStates` entry at the same time we modify `contexts`.

However, then, we have to maintain two separate Map which holds per-partition states, which is bothersome and an error-prone.

So I think we should `reloadRequested` flag in PartitionContext itself. WDYT

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java
```diff
@@ -214,7 +233,9 @@ public List<TopicPartition> partitionsNeedsPause() {
         return contexts.values().stream()
                        .filter(c -> !c.revoking())
                        .filter(c -> !c.paused())
-                       .filter(c -> pausingAll || shouldPartitionPaused(c.pendingTasksCount()))
+                       .filter(c -> pausingAll
+                                    || reloadStates.get(c.topicPartition())
```
ocadaruma, line 237: Additionally, we also need to modify `partitionNeedsResume` to avoid returning partitions that reload is requested, because otherwise, partitions will be resumed immediately so new tasks will be kept feeding so pendingTasks never becomes 0.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java
ta7uw: I got the point.
Having a `reloadRequested` flag in `PartitionContext` seems good.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java
```diff
@@ -53,6 +57,8 @@ public class PartitionContexts implements OffsetsStore, AssignmentStore, Partiti
     private final Map<TopicPartition, PartitionContext> contexts;
 
     private final AtomicBoolean reloadRequested;
```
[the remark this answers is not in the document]
ta7uw: `reloadRequested` is useful for quickly determining if contexts should be reloaded without iteration.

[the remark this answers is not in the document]
ocadaruma: Hm, IMO having duplicated info (with per-partition reload) could be an error prone when we maintain this code in the future, while iterating over partitions unlikely becomes a bottleneck (since we already doing iterating partitions for updating high watermarks, checking partition needs pause/resume) so the benefit to have `reloadRequested` separately is small.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java
```diff
@@ -52,7 +56,7 @@ public class PartitionContexts implements OffsetsStore, AssignmentStore, Partiti
     private final int maxPendingRecords;
     private final Map<TopicPartition, PartitionContext> contexts;
 
-    private final AtomicBoolean reloadRequested;
+    private final ReentrantLock lock;
```
ocadaruma, line 59: Let's rename to something like `propertyReloadLock` for understandability

ocadaruma (commented):
Left few comments but almost looks good.

Also would you mind adding an integration test to check changing the concurrency in the middle of processing doesn't cause problem? (referring RateLimiterTest might be helpful)

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java
```diff
@@ -71,8 +75,12 @@ public PartitionContexts(SubscriptionScope scope, Processors<?> processors) {
                 return;
             }
 
-            if (!reloadRequested.getAndSet(true)) {
+            lock.lock();
+            try {
+                contexts.values().forEach(context -> context.reloadState(true));
```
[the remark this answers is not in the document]
ta7uw: Do you mean having another `ReentrantLock` for updating the `contexts`?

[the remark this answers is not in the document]
ocadaruma: No, use `propertyReloadLock`.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java
```diff
@@ -245,28 +277,37 @@ public void partitionsResumed(List<TopicPartition> partitions) {
...
-            if (totalPendingTasks() > 0) {
-                logger.debug("Waiting pending tasks for property reload.");
+        propertyReloadLock.lock();
+        try {
+            List<TopicPartition> reloadableTopicPartitions = contexts.entrySet()
+                    .stream()
+                    .filter(entry -> entry.getValue().reloadRequested()
+                                     && entry.getValue().pendingTasksCount() == 0)
+                    .map(Entry::getKey)
+                    .collect(Collectors.toList());
```
ocadaruma, line 292: [nits] `toList` is already static-imported so let's omit `Collectors`

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java
```diff
@@ -27,8 +27,12 @@
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Function;
+import java.util.stream.Collectors;
```
ocadaruma, line 35: [nits] There are several unused imports

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java
```diff
@@ -245,28 +277,37 @@ public void partitionsResumed(List<TopicPartition> partitions) {
...
 
-    private void reloadContexts() {
-        // Save current topicPartitions into copy to update contexts map while iterating over this copy.
-        Set<TopicPartition> topicPartitions = new HashSet<>(contexts.keySet());
-
-        logger.info("Start dropping partition contexts");
+    private void reloadContexts(Collection<TopicPartition> topicPartitions) {
+        logger.info("Start dropping partition context({})", topicPartitions);
         removePartition(topicPartitions);
         logger.info("Finished dropping partition contexts. Start recreating partition contexts");
```
ocadaruma, line 312: [nits] This log and this line https://github.com/line/decaton/blob/5659ec7019e70f857f7d39aa86b301fdb69388e5/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java#L316 should also include `topicPartitions`

ocadaruma (commented):
Left last few nits comments!

Also could you fix this comment? https://github.com/line/decaton/blob/5659ec7019e70f857f7d39aa86b301fdb69388e5/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorProperties.java#L77 as reload is now done in per-partition

ocadaruma (approved):
LGTM. Nice!

PR #193, opened 2023-03-08
Introduce a timeout for destroying processors

### LINKED ISSUE #119: Introduce a timeout for destroying processors
## Summary
- Currently, rebalance-listener could block indefinitely in below scenario:
  * Let's say:
      - There are 3 partitions (0,1,2)
      - 3 instances (X,Y,Z)
      - partition-0 is assigned to X
      - `max.poll.interval.ms` is set to long value because sometimes task takes long time to complete (Basically Decaton doesn't block `poll()` by task processing, but there's an exception in rebalance-listener)
  * Then, let's say partition-0's assignment is taken over by Y by reassignment triggered by Z's restart
  * rebalance-listener's `onPartitionAssigned` calls partition-0's `PartitionProcessor#close`, which waits processor unit's executor to terminate indefinitely
  * If long-running task is being executed by a processor, `onPartitionAssigned` will be blocked until the task completes
  * When Z came back to the group, it initiates another rebalance, but it never finishes because X is in stuck at `onPartitionAssigned`, so cannot send `JoinGroup`
  * Meanwhile, all processing on all instances are stuck
- Though this behavior is Decaton's design decision (i.e. "rebalance could block up to `max.poll.interval.ms` at max"), some users use Decaton for processing long-running tasks, and aborting task "ungracefully" is preferable to waiting task indefinitely
- So it might be better to have an option to set timeout on destroying processors

ta7uw:
Motivation:

Described in https://github.com/line/decaton/issues/119

Modifications:

- Add timeout for destroying processors
- Set timeout to `PartitionProcessor#awaitShutdown`

Result:

- Closes #119
- `PartitionProcessor#awaitShutdown` waits up to 1000 ms by default
- Users can easily set this timeout

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorProperties.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorProperties.java
@@ -29,7 +29,6 @@ import com.linecorp.decaton.processor.DecatonProcessor;
 import com.linecorp.decaton.processor.ProcessingContext;
 import com.linecorp.decaton.processor.runtime.internal.AbstractDecatonProperties;
 import com.linecorp.decaton.processor.runtime.internal.OutOfOrderCommitControl;
-import com.linecorp.decaton.processor.runtime.internal.PartitionProcessor;
 import com.linecorp.decaton.processor.runtime.internal.RateLimiter;
 
 /**
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorProperties.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorProperties.java
@@ -29,6 +29,7 @@ import com.linecorp.decaton.processor.DecatonProcessor;
 import com.linecorp.decaton.processor.ProcessingContext;
 import com.linecorp.decaton.processor.runtime.internal.AbstractDecatonProperties;
 import com.linecorp.decaton.processor.runtime.internal.OutOfOrderCommitControl;
+import com.linecorp.decaton.processor.runtime.internal.PartitionProcessor;
 import com.linecorp.decaton.processor.runtime.internal.RateLimiter;
 
 /**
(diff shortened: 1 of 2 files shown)


processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorProperties.java
```diff
@@ -179,6 +180,19 @@ public class ProcessorProperties extends AbstractDecatonProperties {
             PropertyDefinition.define("decaton.deferred.complete.timeout.ms", Long.class, -1L,
                                       v -> v instanceof Long);
 
+    /**
+     * Timeout for destroying processors.
+     * When {@link PartitionProcessor#close()} is called, all processors will be destroyed.
+     * At this time, Decaton waits synchronously for the running tasks to finish until this timeout.
+     * If it is not a problem not to wait for task completion, it is recommended to set this timeout,
```
ocadaruma, line 187: Actually, this feature should be used with caution because:
- When `PartitionProcessor` times out, Decaton will continue other clean-up tasks like destroying partition-scoped DecatonProcessor (https://github.com/line/decaton/blob/v6.2.0/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContext.java#L126), which could cause unexpected behavior in the middle of last `DecatonProcessor#process` which timed-out
- Also, since processor-thread may continue processing last task even after this timeout, the task may be completed at an unexpected timing after a while (e.g. after Decaton already released partition assignment for the task came from)

So this feature should be used only when users want to avoid waiting last task execution at the above risks.

So we should include these implications in the javadoc, and I think it's better not to use word "recommended" here.

ocadaruma (commented):
Overall looks good.

Left few comments

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorProperties.java
```diff
@@ -179,6 +180,22 @@ public class ProcessorProperties extends AbstractDecatonProperties {
             PropertyDefinition.define("decaton.deferred.complete.timeout.ms", Long.class, -1L,
                                       v -> v instanceof Long);
 
+    /**
+     * Timeout for destroying processors.
```
ocadaruma, line 184: I just realized the word `processor` is used in various context also for other properties...

To clarify the effect of this property, how about rewording to `Timeout for processor threads termination` ?

And also renaming the property to `decaton.processor.threads.termination.timeout.ms` or something

ocadaruma (commented):
Left one minor feedback but almost looks good!

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorProperties.java
```diff
@@ -29,6 +29,7 @@
 import com.linecorp.decaton.processor.ProcessingContext;
 import com.linecorp.decaton.processor.runtime.internal.AbstractDecatonProperties;
 import com.linecorp.decaton.processor.runtime.internal.OutOfOrderCommitControl;
+import com.linecorp.decaton.processor.runtime.internal.PartitionProcessor;
```
ocadaruma, line 32: unused import?

ocadaruma (approved):

Issue #187, opened 2023-02-24
Flaky test: CentralDogmaPropertySupplierIntegrationTest


ocadaruma:
https://github.com/line/decaton/actions/runs/4239651470/jobs/7367906297

```
com.linecorp.decaton.centraldogma.CentralDogmaPropertySupplierIntegrationTest > testCDIntegration FAILED
    java.util.concurrent.CompletionException: com.linecorp.armeria.client.UnprocessedRequestException: com.linecorp.armeria.client.endpoint.EndpointSelectionTimeoutException: Failed to select within 3200 ms an endpoint from: HealthCheckedEndpointGroup{endpoints=[Endpoint{127.0.0.1:35307, weight=1000}], numEndpoints=1, candidates=[Endpoint{127.0.0.1:35307, weight=1000}], numCandidates=1, selectionStrategy=class com.linecorp.armeria.client.endpoint.WeightedRoundRobinStrategy, initialized=true}
        at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:332)
        at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:347)
        at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:636)
        at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
        at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2162)
        at com.linecorp.armeria.common.util.UnmodifiableFuture.doCompleteExceptionally(UnmodifiableFuture.java:167)
        at com.linecorp.armeria.common.util.UnmodifiableFuture.lambda$wrap$0(UnmodifiableFuture.java:101)
        at java.base/java.util.concurrent.CompletableFuture.uniHandle(CompletableFuture.java:934)
        at java.base/java.util.concurrent.CompletableFuture$UniHandle.tryFire(CompletableFuture.java:911)
        at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
        at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2162)
        at com.linecorp.armeria.common.stream.DeferredStreamMessage.lambda$delegate$0(DeferredStreamMessage.java:134)
        at java.base/java.util.concurrent.CompletableFuture.uniHandle(CompletableFuture.java:934)
[...]

PR #197, opened 2023-03-29
Improve CentralDogmaPropertySupplierIntegrationTest

thachlp:
Related issue: https://github.com/line/decaton/issues/187
As Javadoc of `CentralDogmaRule` class we should create rules by
```
@ClassRule
public static final CentralDogmaRule rule = new CentralDogmaRule();
```
The problem with `@Rule` is it will start and stop an embedded server for each test, I think it may affect performance tests and can cause the `EndpointSelectionTimeoutException` random fail.

### DIFF
--- a/centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierIntegrationTest.java
+++ b/centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierIntegrationTest.java
@@ -32,7 +32,8 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
-import org.junit.Rule;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
 import org.junit.Test;
 
 import com.fasterxml.jackson.databind.JsonNode;
@@ -53,8 +54,8 @@ import com.linecorp.decaton.processor.runtime.Property;
 
 public class CentralDogmaPropertySupplierIntegrationTest {
 
-    @Rule
-    public CentralDogmaRule centralDogmaRule = new CentralDogmaRule() {
+    @ClassRule
+    public static final CentralDogmaRule centralDogmaRule = new CentralDogmaRule() {
         @Override
         protected void configureHttpClient(WebClientBuilder builder) {
             builder.decorator(RetryingClient.builder(RetryRule.onUnprocessed())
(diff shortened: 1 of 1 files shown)


ocadaruma (commented):
Thanks for the patch!

If we share single CentralDogmaRule instance among tests, doesn't it cause potential race between tests? (e.g. testA create file => testB depends on file's absence)

We should avoid bring such dependencies in tests because it makes tests fragile, so we should fix the root cause of `EndpointSelectionTimeoutException` rather than mitigating the chance by reducing instance creation.

thachlp:
> Thanks for the patch!
> 
> If we share single CentralDogmaRule instance among tests, doesn't it cause potential race between tests? (e.g. testA create file => testB depends on file's absence)
> 
> We should avoid bring such dependencies in tests because it makes tests fragile, so we should fix the root cause of `EndpointSelectionTimeoutException` rather than mitigating the chance by reducing instance creation.

@ocadaruma Thanks for replying to me 😊
As I understand, delcaton uses CentralDogma to store properties, so the test should be the same as the use-case behavior, we start only one centraldogma server, and working on this... I also referenced to centraldogma code to see how they test.
From the issue log, I see it is a random fail, so I just think about the problem, we start/stop the dogma server for each test, so we can not sure if the server is already for access or not (even it started success)

ocadaruma (commented):
H-mm, maybe I still don't quite understand well why this patch could fix the flaky test.

I understand making CentralDogmaRule as ClassRule can reduce the chance of the failure though, still the root cause is not fixed so same failure could happen, right?

>  so the test should be the same as the use-case behavior

Current patch could fail depending on the test execution order. (why currently it succeeds is just "lucky").

In `testFileExist`, it creates a file in the CD project. On the other hand, in `testFileNonExistent` checks the absence of the file.
So if these tests are executed in `testFileExist` => `testFileNonExistent`, `testFileNonExistent` will fail.
(Though JUnit4 determines the test execution order deterministically, we shouldn't write a test that could fail depending on the execution order in general)

So, to get fresh CD instances every time, using `@Rule` is more simple.

thachlp:
> H-mm, maybe I still don't quite understand well why this patch is supposed to fix the flaky test.
> 
> I understand making CentralDogmaRule as ClassRule can reduce the chance of the failure though, still the root cause is not fixed so same failure could happen, right?
> 
> > so the test should be the same as the use-case behavior
> 
> Current patch could fail depending on the test execution order. (why currently it succeeds is just "lucky").
> 
> In `testFileExist`, it creates a file in the CD project. On the other hand, `testFileNonExistent` checks the absence of the file. So if these tests are executed in `testFileExist` => `testFileNonExistent`, `testFileNonExistent` will fail. (Though JUnit4 determines the test execution order deterministically, we shouldn't write a test that could fail depending on the execution order in general)
> 
> So, to get fresh CD instances every time, using `@Rule` is more simple.

Thank you for clear explanation to me @ocadaruma, so I will close this pr and work on another pr (if I can find the root cause), again, thank you

PR #202, opened 2023-05-10
Handle RebalanceInProgressException on sync commit fixes #201

### LINKED ISSUE #201: Subscription could be killed by RebalanceInProgressException
- Decaton swallows some exceptions thrown from offset-commit on partition revocations, as by default, Decaton doesn't consider offset-commit failures as fatal one because process-duplication by commit failures are "normal" behavior which implementer should care about anyways
  * https://github.com/line/decaton/blob/v6.3.1/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorSubscription.java#L83-L84
- After https://github.com/apache/kafka/pull/7312, commitSync now could throw RebalanceInProgressException, which currently Decaton doesn't catch

ocadaruma:
- As described in the issue, RebalanceInProgressException also could be thrown on commitSync, which should not be treated as fatal


kawamuray (approved):
LGTM, thanks for the quick fix.

PR #207, opened 2023-06-24
Make `max.pending.records` property reloadable

KarboniteKream:
This PR updates `PartitionContexts` so the `max.pending.records` property can be reloaded at runtime, similar to the `partition.concurrency` property.

Modifications:
- Extract reload listener in `PartitionContexts` and register it for `CONFIG_MAX_PENDING_RECORDS`
- `PartitionContext` now reads the property value from `PartitionScope`
- Cleanup `ProcessorProperties` Javadoc formatting
- Cleanup some `unchecked` code in tests

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java
@@ -36,6 +36,7 @@ import org.slf4j.LoggerFactory;
 
 import com.linecorp.decaton.processor.runtime.ProcessorProperties;
 import com.linecorp.decaton.processor.runtime.Property;
+import com.linecorp.decaton.processor.runtime.PropertyDefinition;
 import com.linecorp.decaton.processor.runtime.internal.AssignmentManager.AssignmentConfig;
 import com.linecorp.decaton.processor.runtime.internal.AssignmentManager.AssignmentStore;
 import com.linecorp.decaton.processor.runtime.internal.CommitManager.OffsetsStore;
@@ -48,7 +49,7 @@ public class PartitionContexts implements OffsetsStore, AssignmentStore, Partiti
     private final SubscriptionScope scope;
     private final Processors<?> processors;
     private final Property<Long> processingRateProp;
-    private final int maxPendingRecords;
+    private final Property<Integer> maxPendingRecordsProp;
     private final Map<TopicPartition, PartitionContext> contexts;
 
     private final ReentrantLock propertyReloadLock;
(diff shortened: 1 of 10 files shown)


processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java
```diff
@@ -212,13 +199,13 @@ public void updateHighWatermarks() {
     }
 
     private boolean shouldPartitionPaused(int pendingRecords) {
-        return pendingRecords >= maxPendingRecords;
+        return pendingRecords >= maxPendingRecordsProp.value();
```
ocadaruma, line 202: I'm still not 100% sure if there's a race which could cause OutOfOrderCommitControl offset overflow.
I would like to check further.

Please give me some time.

ocadaruma (commented):
Sorry for delaying the review.

Thanks! Left a comment but overall looks good.
Also could you add an integration test in `PropertyReloadRequestTest` to check dynamic max.pending.records reload works?

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java
KarboniteKream: Of course, no rush. I've done some local stress tests and I didn't get any exceptions, but I'm not familiar with the internals, so I don't know which circumstances can trigger this.

ocadaruma: I actually found the race exists in below scenario by using formal methods.(https://gist.github.com/ocadaruma/1f9e61887d3d93e8a73422065b80205f)

initial state:
- `max.pending.records` = 1000
- `max.poll.records` = 100
    * Thus OutOfOrderCommitControl capacity is 1100
- pendingTasks = 999

scenario:

1. [poll()](https://github.com/KarboniteKream/decaton/blob/6460c794e8c18b47ce360efda253495030b60b07/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ConsumeManager.java#L166) returns 100 records and added to OutOfOrderCommitControl
2. [updatePartitionStatus()](https://github.com/KarboniteKream/decaton/blob/6460c794e8c18b47ce360efda253495030b60b07/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java#L218)
    * pendingTasks = 1099
3. [pausePartitions()](https://github.com/KarboniteKream/decaton/blob/6460c794e8c18b47ce360efda253495030b60b07/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java#L229)
    * the partition is determined to be need-pause
4. [resumePartitions() / reloadRequested flag check](https://github.com/KarboniteKream/decaton/blob/6460c794e8c18b47ce360efda253495030b60b07/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java#L241)
    * reload is not requested at this point
5. `max.pending.records` is increased to 2000 dynamically
6. [resumePartitions() / check pending tasks](https://github.com/KarboniteKream/decaton/blob/6460c794e8c18b47ce360efda253495030b60b07/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java#L242)
    * since now pendingTasks < `max.pending.records`, the partition determined to be need-resume
7. [maybeHandlePropertyReload()](https://github.com/KarboniteKream/decaton/blob/6460c794e8c18b47ce360efda253495030b60b07/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/PartitionContexts.java#L266C17-L266C42)
    * no-op since pendingTasks is still non-zero
8. [next poll()](https://github.com/KarboniteKream/decaton/blob/6460c794e8c18b47ce360efda253495030b60b07/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ConsumeManager.java#L166)
    * since the partition is not paused, poll returns some records
    * records are added to OutOfOrderCommitControl (capacity = 1099) => overflow

There might be several solutions though, I think straightforward way is to store `max.pending.records` at the instantiation timing in PartitionContext and comparing pendingTasks against it rather than dynamic property value for pause/resume necessity check.
WDYT?

KarboniteKream: Makes sense, updated

ocadaruma (commented):
Thanks! Left only nits comment

ocadaruma (approved):
LGTM, thanks!

PR #212, opened 2023-09-19
Prevent non-RuntimeException causes completion-leak fixes #211

### LINKED ISSUE #211: Potential completion leak when TaskExtractor#extract throws non-RuntimeException
- As of 7.1.0, Decaton catches any RuntimeExceptions thrown from `TaskExtractor#extract`
  * https://github.com/line/decaton/blob/v7.1.0/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ProcessPipeline.java#L82
- This was to prevent disrupting the processing due to invalid messages
  * In Java, catching only `RuntimeException` was sufficient because any checked-exceptions thrown inside `TaskExtractor#extract` must be handled by developers explicitly, otherwise compilation fails
  * https://github.com/line/decaton/blob/v7.1.0/processor/src/main/java/com/linecorp/decaton/processor/runtime/TaskExtractor.java#L28
- However, when developers implement `TaskExtractor` in Kotlin (or Scala), `TaskExtractor#extract` would throw non-RuntimeException and developers would easily overlook that
  * Since there's no concept of checked/unchecked exceptions in these langs

ocadaruma:
- The problem and the cause is described in #211
- Summary of the fixes:
  * Considering any user-supplied logic could throw non-RuntimeExceptions when it's implemented by non-Java JVM languages, I changed to catch `Exception` instead of `RuntimeException` where Decaton executes user-supplied logic

### DIFF
--- a/centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
+++ b/centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
@@ -123,14 +123,14 @@ public class CentralDogmaPropertySupplier implements PropertySupplier, AutoClose
         child.watch(node -> {
             try {
                 setValue(prop, node);
-            } catch (RuntimeException e) {
+            } catch (Exception e) {
                 logger.warn("Failed to set value updated from CentralDogma for {}", definition.name(), e);
             }
         });
         try {
             JsonNode node = child.initialValueFuture().join().value(); //doesn't fail since it's a child watcher
             setValue(prop, node);
-        } catch (RuntimeException e) {
+        } catch (Exception e) {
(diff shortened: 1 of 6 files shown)


centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
```diff
@@ -123,14 +123,14 @@ public <T> Optional<Property<T>> getProperty(PropertyDefinition<T> definition) {
         child.watch(node -> {
             try {
                 setValue(prop, node);
-            } catch (RuntimeException e) {
+            } catch (Exception e) {
```
kawamuray, line 126: Maybe it's better to leave a short comment, or it could be reverted back to RE by "code cleaners"...

kawamuray (changes requested):

kawamuray (approved):

PR #214, opened 2023-10-03
Drop Java8 support, upgrade gradle and junit

kawamuray:
* Drop java 8 and support java 20
* Gradle version to 8.2
* junit version to 5.10.0
* along with other testlibs

Will be required to use higher java version, but should be worth doing by itself.


kawamuray:
Apparently integrationTest doesn't run correctly. Investigating.

ocadaruma:
jdk8 test is failing?

kawamuray:
So dropping java8 support is a consequence of upgrading mockito lib. While I don't think newer mockito lib itself justifies to drop java 8 support, I guess many libs will soon drop support of 8 so we're gonna need to do it anyways.
If anyone's having trouble for losing java 8 support, we're happy to re-consider how can we keep support java 8 for a bit more while.

client/src/test/java/com/linecorp/decaton/client/internal/DecatonClientImplTest.java
```diff
@@ -31,28 +31,28 @@
...
+import org.mockito.junit.jupiter.MockitoSettings;
+import org.mockito.quality.Strictness;
 
 import com.linecorp.decaton.client.DecatonClient.TaskMetadata;
 import com.linecorp.decaton.protobuf.ProtocolBuffersSerializer;
 import com.linecorp.decaton.protocol.Decaton.DecatonTaskRequest;
 import com.linecorp.decaton.protocol.Sample.HelloTask;
 
+@ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
```
ocadaruma, line 50: Do we need `LENIENT` here?

ocadaruma (commented):
Thanks for your work!

Left only minor comments

client/src/test/java/com/linecorp/decaton/client/internal/DecatonClientImplTest.java
kawamuray: Otherwise it raises `UnecessaryMockException`.

brave/src/it/java/com/linecorp/decaton/processor/runtime/BraveTracingTest.java
```diff
@@ -71,24 +71,25 @@ public void doAssert() {
...
 
-    @After
+    @AfterEach
     public void tearDown() {
         rule.admin().deleteTopics(true, retryTopic);
     }
 
-    @Test(timeout = 30000)
+    @Test
+    @Timeout(30000)
```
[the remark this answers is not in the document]
kawamuray: agh... really, nice catch.

client/src/test/java/com/linecorp/decaton/client/internal/DecatonClientImplTest.java
ocadaruma: Let's remove unnecessary mocks then

brave/src/it/java/com/linecorp/decaton/processor/runtime/BraveTracingTest.java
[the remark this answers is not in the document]
kawamuray: fixed up.

client/src/test/java/com/linecorp/decaton/client/internal/DecatonClientImplTest.java
ocadaruma: Hmm, fair enough.

Agree with the strategy.

ocadaruma (approved):

ocadaruma:
Great work!

mauhiz:
I think some of your most fervent users still build with Java8 :/

kawamuray:
oops, I'll reconsider then :o

PR #218, opened 2023-11-22
Add delivery latency and scheduled delay metrics

### LINKED ISSUE #208: Expose e2e delivery latency metrics
- Each task has timestamp metadata so it should be trivial

Yang-33:
The Decaton client adds timestamps to tasks when they are produced, as well as the scheduled execution time. As mentioned in #208, this allows us to output metrics on the time taken from producing to consuming a task. This can be beneficial for monitoring Decaton processors that require low latency, and for creating alerts.

Additionally, metrics on the delay from the scheduled time are particularly useful in determining whether schedules are being executed in retry topic as expected. If there is a significant delay, Decaton users may realize they need increase throughput somehow.

Resolve #208

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
@@ -147,17 +147,11 @@ public class Metrics {
     public class TaskMetrics extends AbstractMetrics {
         public final Timer tasksDeliveryLatency =
                 meter(() -> Timer.builder("tasks.delivery.latency")
-                        .description("The latency between the time the task is produced and the time the task is consumed. "
-                                + "This metric depends on the task's `timestampMillis` field, and "
-                                + "it might not represent the actual end-to-end latency depending on how `timestampMillis` is constructed for the task.")
-                        .tags(availableTags.partitionScope())
-                        .register(registry));
-        public final Timer tasksDeliveryLatency =
-                meter(() -> Timer.builder("tasks.delivery.latency")
-                                 .description("The latency between the time the task is produced and the time the task is consumed")
+                                 .description("The latency between the time the task is produced and the time the task is consumed. "
+                                         + "This metric depends on the task's `timestampMillis` field, and it might not represent "
+                                         + "the actual end-to-end latency depending on how `timestampMillis` is constructed for the task.")
                                  .tags(availableTags.partitionScope())
                                  .register(registry));
-
         public final Timer tasksScheduledDelay =
                 meter(() -> Timer.builder("tasks.scheduled.delay")
                                  .description("The delay between the scheduled time and the time the task is consumed")
--- a/processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
@@ -147,6 +147,13 @@ public class Metrics {
     public class TaskMetrics extends AbstractMetrics {
         public final Timer tasksDeliveryLatency =
                 meter(() -> Timer.builder("tasks.delivery.latency")
+                        .description("The latency between the time the task is produced and the time the task is consumed. "
+                                + "This metric depends on the task's `timestampMillis` field, and "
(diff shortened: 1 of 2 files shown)


processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
```diff
@@ -145,6 +145,18 @@ public void close() {
     }
 
     public class TaskMetrics extends AbstractMetrics {
+        public final Timer tasksDeliveryLatency =
+                meter(() -> Timer.builder("tasks.delivery.latency")
+                                 .description("The latency between the time the task is produced and the time the task is consumed")
```
kawamuray, line 150: We should mention that it might not represent the actual e2e latency depending on how is `timestampMillis` constructed for the task, particularly when `TaskExtractor` interface is used, user might not set appropriate `timestampMillis` (which is OK if user is aware of that, but want to avoid misleading users by letting them to assume e2e latency is somehow way shorter than actually is).

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ProcessPipeline.java
```diff
@@ -124,6 +125,19 @@ Completion process(TaskRequest request, DecatonTask<T> task) throws InterruptedE
                 new ProcessingContextImpl<>(scope.subscriptionId(), request, task, processors, retryProcessor,
                                             scope.props());
 
+        if (task.metadata().timestampMillis() > 0) {
```
kawamuray, line 128: Can you put this logic here instead? https://github.com/line/decaton/pull/218/files#diff-4df160cb6c95ec9c8815d2af1dda8a9ebe841c74c4e094bd3ef8e75074dd0af5R100

Just because I think we want to keep this `process` method to have lines that are essential to execute the stage of executing `DecatonProcessor#process`.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ProcessPipeline.java
```diff
@@ -124,6 +125,19 @@ Completion process(TaskRequest request, DecatonTask<T> task) throws InterruptedE
...
+            taskMetrics.tasksDeliveryLatency.record(
+                    System.currentTimeMillis() - task.metadata().timestampMillis(),
```
kawamuray, line 130: Please assign this to a variable and reuse for the below 2nd call place.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ProcessPipeline.java
kawamuray, line 128: Maybe we better to filter-out `currentTimeMillis() - timestampMillis() < 0` cases to avoid counting obviously broken timestamp value? (and likewise for the below `tasksScheduledDelay` case?)

kawamuray (changes requested):
Thanks for your PR! Generally sounds good, just a few comments left.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ProcessPipeline.java
Yang-33: As https://github.com/line/decaton/pull/218#discussion_r1403968994 said, indeed the filter-out is necessary because we can freely modify the metadata timestamp. Thank you!

processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java
Yang-33: That's indeed true. This metric heavily relies on `timestampMillis`. Based on your review comment, I've added a note to the description of the metrics.

I have a question. Should this also include in the description about the case where an inappropriate `timestampMillis` is set when using TaskExtractor or decaton client's `put`? Or is a comment sufficient?

kawamuray: I think the current statement is sufficient.

kawamuray (approved):
LGTM, thanks!

PR #219, opened 2023-11-24
Some updates and improvements for the benchmark module

kawamuray:
* Upgrade async-profiler version (to support future JVM versions) 
* Metrics accessibility for benchmark execution
* Get rid of thread local for recordings

### DIFF
--- a/benchmark/src/main/java/com/linecorp/decaton/benchmark/Recording.java
+++ b/benchmark/src/main/java/com/linecorp/decaton/benchmark/Recording.java
@@ -17,11 +17,10 @@
 package com.linecorp.decaton.benchmark;
 
 import java.time.Duration;
-import java.util.ArrayList;
-import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 
 import com.linecorp.decaton.benchmark.BenchmarkResult.Performance.Durations;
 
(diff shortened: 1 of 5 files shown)


benchmark/src/main/java/com/linecorp/decaton/benchmark/Recording.java
```diff
@@ -17,33 +17,33 @@
...
+        private final AtomicLong maxDeliveryTime = new AtomicLong();
+        private final AtomicLong totalDeliveryTime = new AtomicLong();
+        private final AtomicLong processCount = new AtomicLong();
 
         void process(Task task, boolean warmup) {
             if (!warmup) {
                 long deliveryLatency = System.currentTimeMillis() - task.getProducedTime();
-                if (deliveryLatency > maxDeliveryTime) {
-                    maxDeliveryTime = deliveryLatency;
+                while (true) {
```
ocadaruma, line 39: Can't we simplify to `maxDeliveryTime.updateAndGet(current -> Math.max(current, deliveryLatency))` ?

ocadaruma (commented):
Thanks! Left minor comment.

benchmark/src/main/java/com/linecorp/decaton/benchmark/Recording.java
ocadaruma: We found this short circuit `deliveryLatency <= currentMax ||` is important for performance since compareAndSet is certainly time consuming.

Okay then LGTM with current code

ocadaruma (approved):

benchmark/src/main/java/com/linecorp/decaton/benchmark/Recording.java
kawamuray: For the record:
```java
@BenchmarkMode(Mode.Throughput)
@Fork(1)
@Warmup(iterations = 3)
@Measurement(iterations = 3)
@Threads(20)
public class AtomicOperationBenchmark {
    @State(Scope.Benchmark)
    public static class BmState {
        private final AtomicLong atomic = new AtomicLong();
    }

    @Benchmark
    public void loop(BmState state) {
        long value = 100;
        while (true) {
            long current = state.atomic.get();
            if (value <= current || state.atomic.compareAndSet(current, value)) {
                break;
            }
        }
    }

    @Benchmark
    public void updateAndGet(BmState state) {
        long value = 100;
        state.atomic.updateAndGet(current -> Math.max(current, value));
    }

}
```

```
Result "com.linecorp.decaton.processor.runtime.AtomicOperationBenchmark.loop":                                                                
 4031830803.542 ±(99.9%) 356905983.395 ops/s [Average]                                                                           
 (min, avg, max) = (4009246675.155, 4031830803.542, 4043555863.590), stdev = 19563221.275                                                          
 CI (99.9%): [3674924820.147, 4388736786.936] (assumes normal distribution)

Result "com.linecorp.decaton.processor.runtime.AtomicOperationBenchmark.updateAndGet":                                                                                                                      
  23741034.872 ±(99.9%) 1267965.111 ops/s [Average]                                                                                                                                                         
  (min, avg, max) = (23690787.656, 23741034.872, 23820351.321), stdev = 69501.446                                                                                                                           
  CI (99.9%): [22473069.762, 25008999.983] (assumes normal distribution)   
```

kawamuray: Didn't know that method btw, nice to know, thanks.

PR #222, opened 2023-12-13
Cache CD dynamic properties

### LINKED ISSUE #221: CentralDogmaPropertySupplier may leak watchers
`CentralDogmaPropertySupplier.getProperty` will make a new child watcher on each call; but does not cache them in any way.
This is probably intended, but the documentation for this method does not make it explicit enough that calling it repeatedly leaks memory.

In order to prevent accidental memory leaks, we should keep a cache of property definition name to CD-supplied property

mauhiz:
Fixes #221

### DIFF
--- a/centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
+++ b/centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
@@ -18,6 +18,8 @@ package com.linecorp.decaton.centraldogma;
 
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
@@ -43,6 +45,9 @@ import com.linecorp.decaton.processor.runtime.ProcessorProperties;
 import com.linecorp.decaton.processor.runtime.Property;
 import com.linecorp.decaton.processor.runtime.PropertyDefinition;
 import com.linecorp.decaton.processor.runtime.PropertySupplier;
+
+import lombok.AccessLevel;
+import lombok.Getter;
 
 /**
  * A {@link PropertySupplier} implementation with Central Dogma backend.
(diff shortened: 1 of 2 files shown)


centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
```diff
@@ -118,32 +126,43 @@ public <T> Optional<Property<T>> getProperty(PropertyDefinition<T> definition) {
...
-            JsonNode node = child.initialValueFuture().join().value(); //doesn't fail since it's a child watcher
-            setValue(prop, node);
-        } catch (Exception e) {
-            // Catching Exception instead of RuntimeException, since
-            // Kotlin-implemented DynamicProperty would throw checked exceptions
-            logger.warn("Failed to set initial value from CentralDogma for {}", definition.name(), e);
-        }
 
-        return Optional.of(prop);
+            return Optional.of(prop);
```
ocadaruma, line 152: Why not storing `DynamicProperty<?>` directly instead of wrapping with `Optional` ?

ocadaruma (commented):
Good catching, thank you for the patch!

Left minor question.

centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
mauhiz: Ah, no good reason. Fixed

centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierIntegrationTest.java
```diff
@@ -119,6 +120,11 @@ public void testCDIntegration() throws InterruptedException {
 
         latch.await();
         assertEquals(20, prop.value().intValue());
+
+        IntStream.range(0, 10000)
+                 .mapToObj(i -> CONFIG_PARTITION_CONCURRENCY)
+                 .map(supplier::getProperty);
+        assertEquals(1, supplier.getCachedProperties().size());
```
ocadaruma, line 127: To check cache is working properly, I think it's better all property returned from `supplier.getProperty` are same reference, rather than checking the internal hashmap's content

ocadaruma (commented):
Thanks for the update. left one comment

centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierIntegrationTest.java
mauhiz: Fair enough. Fixed

ocadaruma (approved):
Thanks LGTM!

PR #225, opened 2024-03-06
Ensure decaton works with older jdk

ocadaruma:
- Now Decaton compiles only on jdk >= 21 (by https://github.com/line/decaton/pull/224) but since there are many users who still use jdk 8,11,17, we still should run tests on these versions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
         java: [8, 11, 17, 21]
     steps:
     - uses: actions/checkout@v2
-    - name: Setup java for compile
+    - name: Setup java
       uses: actions/setup-java@v4
       with:
         distribution: temurin
@@ -26,7 +26,9 @@ jobs:
     - name: Execute test
       uses: eskatos/gradle-command-action@v1
       with:
+        # Java is installed on JAVA_HOME_{java major version}_X64
+        # refs: https://github.com/actions/setup-java/tree/v4.1.0?tab=readme-ov-file#install-multiple-jdks
         arguments: |
           -Ptest.java.major.version=${{ matrix.java }}
-          -Porg.gradle.java.installations.fromEnv=JAVA_HOME_${{ matrix.java }}_x64
+          -Porg.gradle.java.installations.fromEnv=JAVA_HOME_${{ matrix.java }}_X64
           build jmhJar integrationTest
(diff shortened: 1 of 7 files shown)


client/src/test/java/com/linecorp/decaton/client/DecatonClientTest.java
```diff
@@ -39,8 +39,7 @@
 
 @ExtendWith(MockitoExtension.class)
 public class DecatonClientTest {
-    @Spy
-    private final DecatonClient<HelloTask> decaton = new DecatonClient<HelloTask>() {
+    private static class NoopClient implements DecatonClient<HelloTask> {
```
ocadaruma, line 42: Spying anonymous class fails on JDK8 with reflection error
https://github.com/line/decaton/actions/runs/8174863099/job/22350646261
image

I didn't want to dig into the detail as we likely drop JDK8-support in not-so-far-future so I just rewrote it with static class for now.

.github/workflows/ci.yml
```diff
@@ -12,14 +12,23 @@ jobs:
...
     - uses: actions/checkout@v2
     - name: Setup java
-      uses: actions/setup-java@v1
+      uses: actions/setup-java@v4
       with:
-        java-version: ${{ matrix.java }}
+        distribution: temurin
+        java-version: |
+          ${{ matrix.java }}
+          21
```
kawamuray, line 24: Is this intentionally left or just forgot to remove line?

processor/src/it/java/com/linecorp/decaton/processor/VThreadCoreFunctionalityTest.java
```diff
@@ -33,6 +35,7 @@
 import com.linecorp.decaton.testing.RandomExtension;
 import com.linecorp.decaton.testing.processor.ProcessorTestSuite;
 
+@EnabledForJreRange(min = JRE.JAVA_21)
```
kawamuray, line 38: Oh... there's such a convenient annotation existed... !

kawamuray (changes requested):
just one minor question

.github/workflows/ci.yml
ocadaruma: Intentionally. We need to install 21 always to build the project, regardless test-jdk version.
Last one will be used as the global java version (refs: https://github.com/actions/setup-java?tab=readme-ov-file#install-multiple-jdks)

kawamuray: ah-ok. now understand how it works

kawamuray (approved):
LGTM, thanks for the quick follow-up!

PR #228, opened 2024-03-07
Fix invalid json of benchmark output

ocadaruma:
- Now performance dashboard doesn't work because latest run produces invalid JSON output
    * https://line.github.io/decaton/commit-data/a3f14197b71c65a7aa8d83b78d5c9756dbf9ca89/tasks_100k_latency_10ms_concurrency_20-benchmark.json
- This is because current version of async-profiler output `Profiling started` on stdout and there seems to be no way to disable this
- To address this, I added `--file` option to the benchmark module to prevent async-profiler output and benchmark output are mixed

### DIFF
--- a/benchmark/src/main/java/com/linecorp/decaton/benchmark/Main.java
+++ b/benchmark/src/main/java/com/linecorp/decaton/benchmark/Main.java
@@ -18,6 +18,7 @@ package com.linecorp.decaton.benchmark;
 
 import static java.util.Collections.emptyList;
 
+import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


ocadaruma:
The invalid json broke performance dashboard so we will need to manually fix it and commit to `gh-pages` branch

benchmark/src/main/java/com/linecorp/decaton/benchmark/Main.java
```diff
@@ -95,6 +96,9 @@ public final class Main implements Callable<Integer> {
             defaultValue = "text")
     private String resultFormat;
 
+    @Option(names = "--file", description = "Result file to write output to. If not specified, output to stdout")
```
kawamuray, line 99: ogh.. thanks again for another follow-up. Just a nit, would it be bertter to name this param `--output` following the convention of most command-line utilities?

kawamuray (changes requested):

kawamuray (approved):

PR #230, opened 2024-03-26
Some fixes for benchmark

kawamuray:
Follow up for #224 

It turns out that the throughput degradation after the above PR is by the newly added `-Xcomp` option in debm.sh
https://line.github.io/decaton/


ocadaruma (approved):
Interesting...

Nice catch!

PR #231, opened 2024-03-27
Add document for subpartition runtime

kawamuray:
Also piggy-back benchmark addition for the virtual thread runtime.


docs/runtime.adoc
```diff
@@ -0,0 +1,94 @@
...
+  --format json \
+  --tasks 10000 \
+  --warmup 100000 \
+  --simulate-latency=4 \
+  --latency-count=5 \
+  --param=decaton.partition.concurrency=$conc \
+  --param=decaton.subpartition.runtime=THREAD_POOL
+----
+
+image::runtime-benchmark-conc-peak.png[]
```
ocadaruma, line 91: Though I understand that the focus of this appendix is about throughput-cap, I guess this graph might make confusing impression because 30000 with 20ms (4ms * 5) latency is still higher than vthread throughput shown in above graph. (So people may think "Got it, threadpool-runtime throughput will be capped by 30000 in 20ms-latency scenario. But vthread-runtime can't achieve 30000 in the first place....")

Isn't it better to show throughput cap with more high simulated latency like 100ms?

ocadaruma (commented):
Nice. Left one feedback

docs/runtime.adoc
ocadaruma: Hm, changed my mind.
Maybe I'm worrying too much. LGTM

ocadaruma (approved):

PR #234, opened 2024-04-11
Fix ResourceTracker in benchmark to work fine with VIRTUAL_THREAD

kawamuray:


benchmark/src/main/java/com/linecorp/decaton/benchmark/DecatonRunner.java
```diff
@@ -113,22 +125,69 @@ public String get(String key) {
...
+        // in the execution context, as in VirtualThread gets created and discarded for all tasks separately.
+        // Instead we need to observe resource usage based on the carrier thread (platform thread) side, while
+        // stdlib doesn't provide a way to resolve the carrier thread's ID.
+        ThreadMXBean threadMxBean = ResourceTracker.getSunThreadMxBean();
+        ThreadInfo[] threadInfos = threadMxBean.dumpAllThreads(true, true);
+        for (ThreadInfo threadInfo : threadInfos) {
+            String name = threadInfo.getThreadName();
+            log.debug("Tracking target check for thread name: {}", name);
+            if (name.startsWith("DecatonSubscriptionThread-") || name.startsWith("PartitionProcessorThread-") ||
+                subPartitionRuntime == SubPartitionRuntime.VIRTUAL_THREAD && name.startsWith("ForkJoinPool-")) {
```
ocadaruma, line 184: So if we have another fork join pool (though currently we don't) in benchmark runtime or somewhere, it may be miscounted as the resource usage.

I think it's acceptable compromise for now though, can we identify virtual thread runtime's pool more precisely somehow? (e.g. by customizing carrier thread's thread factory?)

benchmark/src/main/java/com/linecorp/decaton/benchmark/DecatonRunner.java
```diff
@@ -113,22 +125,69 @@ public String get(String key) {
...
+     * VirtualThread instances directly.
+     */
+    private void maybeSetupForkJoinPoolDrip() {
+        if (subPartitionRuntime != SubPartitionRuntime.VIRTUAL_THREAD) {
+            return;
+        }
+        ScheduledExecutorService executor = Executors.newScheduledThreadPool(
+                1, Thread.ofPlatform().daemon().factory());
+        executor.scheduleAtFixedRate(() -> {
+            for (int i = 0; i < Runtime.getRuntime().availableProcessors(); i++) {
```
ocadaruma, line 160: Does submitting tasks up to `availableProcessors` count ensure keeping all ForkJoinPool worker thread to keep alive?

ocadaruma (commented):
Thank you for the fix.
Left two questions.

benchmark/src/main/java/com/linecorp/decaton/benchmark/DecatonRunner.java
kawamuray: Yeah it worked based on my observation.

benchmark/src/main/java/com/linecorp/decaton/benchmark/DecatonRunner.java
kawamuray: None without any sorts of dirty work :(

ocadaruma: I see, thanks. The only way I can imagine (substituting VirtualThread's scheduler by accessing package-private class) is also dirty so let's do in current way for now

ocadaruma (approved):

PR #235, opened 2024-05-16
Use TopicPartition.topic for metrics

tadashiya:
- I noticed that the metrics always use `scope.topic()` after #224.
- I think it's useful to keep metrics separate between normal topic and retry topic. This change will use `scope.topicPartition().topic()` to use the retry topic name for metrics when handling the retry topic.

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/AbstractSubPartitions.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/AbstractSubPartitions.java
@@ -58,11 +58,11 @@ public abstract class AbstractSubPartitions implements SubPartitions {
         shutdownTimeoutMillis = scope.props().get(
                 ProcessorProperties.CONFIG_PROCESSOR_THREADS_TERMINATION_TIMEOUT_MS);
         rateLimiter = new DynamicRateLimiter(rateProperty(scope));
-        TopicPartition tp = scope.topicPartition();
+        TopicPartition topicPartition = scope.topicPartition();
         Metrics metrics = Metrics.withTags(
                 "subscription", scope.subscriptionId(),
-                "topic", tp.topic(),
-                "partition", String.valueOf(tp.partition()));
+                "topic", topicPartition.topic(),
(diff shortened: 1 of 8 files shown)


processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/AbstractSubPartitions.java
```diff
@@ -59,7 +59,7 @@ protected AbstractSubPartitions(PartitionScope scope, Processors<?> processors)
         rateLimiter = new DynamicRateLimiter(rateProperty(scope));
         Metrics metrics = Metrics.withTags(
                 "subscription", scope.subscriptionId(),
-                "topic", scope.topic(),
+                "topic", scope.topicPartition().topic(),
```
kawamuray, line 62: in addition to these changes I wanna make the following changes to ensure we won't repeat the same mistake in future. Are you happy to follow these up in this PR? Otherwise I'll cover that after merging this PR.

* Let's use variable to assign `scope.topicPartition()` return and use it instead to avoid unnecessary object allocation.
* Rename `SubscriptionScope.topic` to be `SubscriptionScope.originTopic` to distinguish it w/ PartitionScope.topicPartition

kawamuray (changes requested):
Oh very nice catch...!  I overlooked that these two topic references may refer different topics in case of retry and indeed it should refer to the currently consuming topic rather than the origin subscription topic.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/AbstractSubPartitions.java
tadashiya: Thank you for your review. I see, I'll update them as well.

kawamuray (approved):
LGTM, thanks!

kawamuray:
This has been released as 8.0.1. Thanks for your contribution :) https://github.com/line/decaton/releases/tag/v8.0.1

tadashiya:
OK, thanks for your help.

PR #237, opened 2024-06-07
Fix potential key exposure in PKQ

ocadaruma:
## Summary

- Depending on data, key might contain sensitive information so logging it is not preferable

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/QuotaApplierImpl.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/QuotaApplierImpl.java
@@ -17,7 +17,6 @@
 package com.linecorp.decaton.processor.runtime.internal;
 
 import java.time.Duration;
-import java.util.Arrays;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


kawamuray (approved):

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/QuotaApplierImpl.java
```diff
@@ -71,7 +70,7 @@ public boolean apply(ConsumerRecord<byte[], byte[]> record,
             try {
                 topic = callback.apply(record, quotaUsage.metrics());
             } catch (Exception e) {
-                log.error("Exception thrown from the quota callback for key: {}", Arrays.toString(record.key()), e);
+                log.error("Exception thrown from the quota callback", e);
```
mauhiz, line 73: don't we want to log the record offset maybe instead?

ocadaruma: It depends on the QuotaCallback implementation though, IMO logging the exception is enough to tell what happened.
Users can add logging inside QuotaCallback implementation if necessary.

kawamuray: in the first place the user can optionally log more information through enabling format with MDC values?

PR #239, opened 2024-08-14
Update docs and examples for 9.0.0

ocadaruma:
- #238 brings a breaking change, which requires to update docs

### DIFF
--- a/docs/consuming-any-topic.adoc
+++ b/docs/consuming-any-topic.adoc
@@ -1,20 +1,30 @@
 Consuming Arbitrary Topic
 =========================
-:base_version: 8.0.0
+:base_version: 9.0.0
 :modules: common,protocol,processor
 
-This document guides you how to consume and process topics containing records not consists of Decaton's protocol (not produced by DecatonClient) using Decaton processors.
+This document covers an advanced usage of Decaton processor when you process topics containing records not produced by DecatonClient.
 
-By default, Decaton assumes messages are serialized as `DecatonTaskRequest` defined in link:../protocol/src/main/proto/decaton.proto[decaton.proto], and `DecatonProcessor` extracts the task from its `serialized_task` field.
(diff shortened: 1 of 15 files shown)


docs/consuming-any-topic.adoc
```diff
@@ -1,20 +1,30 @@
...
 :modules: common,protocol,processor
 
-This document guides you how to consume and process topics containing records not consists of Decaton's protocol (not produced by DecatonClient) using Decaton processors.
+This document covers an advanced usage of Decaton processor when you process topics containing records not produced by DecatonClient.
 
-By default, Decaton assumes messages are serialized as `DecatonTaskRequest` defined in link:../protocol/src/main/proto/decaton.proto[decaton.proto], and `DecatonProcessor` extracts the task from its `serialized_task` field.
-But Decaton has the capability to consume arbitrary topics other than topics consisting records of Decaton protocol.
+[NOTE]
+====
+* From Decaton 9.0.0, DecatonClient no longer wraps tasks with DecatonTaskRequest protobuf message.
```
kawamuray, line 10: Hm, maybe we can emphasize that most users has no motivation to refer this doc anymore unless they want to do something very niche in the first line of this document, with *bold* letters.
I know it's written below but my point is not as one of these listings but as the headline of the document.

kawamuray (changes requested):

docs/consuming-any-topic.adoc
ocadaruma: Sorry, I recalled that we still need to use `TaskExtractor` for arbitrary topic which is not produced by DecatonClient, because `ProcessorsBuilder.consuming(String topic, Deserializer<T> deserializer)` assumes task is serialized as `DecatonTaskRequest` when `dt_meta` header is absent.

kawamuray (approved):

PR #241, opened 2024-08-15
Make Decaton can consume any topic with deserializer

ocadaruma:
## Motivation
- As of Decaton 8.0.1, the most common way to use Decaton was to use DecatonClient as the producer and DecatonProcessor as the consumer, with implementing value Serializer/Deserializer respectively.
- After #238, DecatonClient no longer wraps tasks with `DecatonTaskRequest` protobuf, which means records produced by DecatonClient and ones produced by non-DecatonClient producers (serialized by Kafka's serializer) have identical serialized bytes.
  * Then why not making `Deserializer` support consuming non-DecatonClient topics, without forcing users to implement `TaskExtractor` (which is bit bothersome than just a deserializer) ?

## Summary of changes
- Update documentations and examples
- `ProcessorsBuilder#consuming(String topic, Deserializer<T> deserializer)` now deserializes messages bytes directly even when task-metadata header is missing. **This is a breaking change**
  * Previous behavior when task-metadata header is missing was to unwrap as `DecatonTaskRequest` pb first, then deserialize serialized task in it. This was for `<= 8.0.0 to 9.0.0` migration where old DecatonClient might produce records during the upgrade.
      * Since "records are produced by old client / or records are produced by non-Decaton client" is not distinguishable from Decaton, we add new `decaton.parse.as.legacy.format.when.header.missing` property to control the behavior for graceful migration.
  * With regards to this, when task-metadata header is missing, we need to fill reasonable default of task metadata.
      * The only field must be filled might be `TaskMetadata#timestampMillis`.
          - e.g. to calculate delivery latency correctly
      * To fill `timestampMillis`, we add `ConsumedRecord#recordTimestampMillis`, which can be retrieved from record itself

## Expected upgrade procedure from <= 8.0.1 to 9.0.0
- Enable both `decaton.retry.task.as.legacy.format` and `decaton.parse.as.legacy.format.when.header.missing`
- Upgrade all processor instances
- Upgrade all decaton client
- Disable `decaton.retry.task.as.legacy.format`
- Disable `decaton.parse.as.legacy.format.when.header.missing`

I'll describe the detail in 9.0.0 release note

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorProperties.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorProperties.java
@@ -24,6 +24,7 @@ import java.util.Map;
 
 import org.slf4j.MDC;
 
+import com.linecorp.decaton.common.Deserializer;
 import com.linecorp.decaton.processor.Completion;
 import com.linecorp.decaton.processor.DecatonProcessor;
 import com.linecorp.decaton.processor.ProcessingContext;
@@ -226,17 +227,31 @@ public class ProcessorProperties extends AbstractDecatonProperties {
                                       Long.MAX_VALUE, v -> v instanceof Long && (Long) v >= 0);
 
     /**
-     * Controls whether to produce retry tasks with task metadata as headers, instead of as deprecated
-     * {@link DecatonTaskRequest} format.
+     * Controls whether to produce retry tasks in deprecated {@link DecatonTaskRequest} format.
      * <p>
-     * <b>CAUTION!!! YOU MAY NEED TO SET THIS TO FALSE WHEN YOU UPGRADE FROM 8.0.1 OR EARLIER</b>
+     * <b>CAUTION!!! YOU MAY NEED TO SET THIS TO TRUE WHEN YOU UPGRADE FROM 8.0.1 OR EARLIER</b>
      * <p>
      * Please read <a href="https://github.com/line/decaton/releases/tag/v9.0.0">Decaton 9.0.0 Release Note</a> carefully.
      * <p>
      * Reloadable: yes
      */
-    public static final PropertyDefinition<Boolean> CONFIG_TASK_METADATA_AS_HEADER =
-            PropertyDefinition.define("decaton.task.metadata.as.header", Boolean.class, true,
+    public static final PropertyDefinition<Boolean> CONFIG_RETRY_TASK_AS_LEGACY_FORMAT =
+            PropertyDefinition.define("decaton.retry.task.as.legacy.format", Boolean.class, false,
+                                      v -> v instanceof Boolean);
+
+    /**
+     * Controls whether to parse records as {@link DecatonTaskRequest} format when task metadata header is missing
+     * when {@link Deserializer} is used, instead of parsing task directly with the deserializer and
+     * fill reasonably-default task metadata.
+     * <p>
+     * <b>CAUTION!!! YOU MAY NEED TO SET THIS TO TRUE WHEN YOU UPGRADE FROM 8.0.1 OR EARLIER</b>
+     * <p>
+     * Please read <a href="https://github.com/line/decaton/releases/tag/v9.0.0">Decaton 9.0.0 Release Note</a> carefully.
+     * <p>
+     * Reloadable: yes
+     */
+    public static final PropertyDefinition<Boolean> CONFIG_PARSE_AS_LEGACY_FORMAT_WHEN_HEADER_MISSING =
+            PropertyDefinition.define("decaton.parse.as.legacy.format.when.header.missing", Boolean.class, false,
                                       v -> v instanceof Boolean);
 
     public static final List<PropertyDefinition<?>> PROPERTY_DEFINITIONS =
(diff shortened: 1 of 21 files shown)


processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorProperties.java
```diff
@@ -226,17 +227,31 @@ public class ProcessorProperties extends AbstractDecatonProperties {
...
+     * <b>CAUTION!!! YOU MAY NEED TO SET THIS TO TRUE WHEN YOU UPGRADE FROM 8.0.1 OR EARLIER</b>
      * <p>
      * Please read <a href="https://github.com/line/decaton/releases/tag/v9.0.0">Decaton 9.0.0 Release Note</a> carefully.
      * <p>
      * Reloadable: yes
      */
-    public static final PropertyDefinition<Boolean> CONFIG_TASK_METADATA_AS_HEADER =
-            PropertyDefinition.define("decaton.task.metadata.as.header", Boolean.class, true,
+    public static final PropertyDefinition<Boolean> CONFIG_RETRY_TASK_AS_LEGACY_FORMAT =
+            PropertyDefinition.define("decaton.retry.task.as.legacy.format", Boolean.class, false,
```
ocadaruma, line 239: I renamed the property and flipped the meaning, because
- For `decaton.parse.as.legacy.format.when.header.missing`, I couldn't figure out good "enabled-by-default" naming.
- Then, for `decaton.task.metadata.as.header`, during migration, "Disable one and enable one" might be too confusing

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/DefaultTaskExtractor.java
```diff
@@ -31,8 +31,13 @@
 
 @RequiredArgsConstructor
 public class DefaultTaskExtractor<T> implements TaskExtractor<T> {
+    private static final ThreadLocal<Boolean> parseAsLegacyWhenHeaderMissing = ThreadLocal.withInitial(() -> false);
```
kawamuray, line 34: h-m this might be a bit too hacky, given that we would have to maintain this compatibility mode for certain period of time isn't it?

Can we do it like;

1. delay `DefaultTaskExctractor` instantiation to the time calling `ProcessorsBuilder#build`
2. so that we can instantiate it in `SubscriptionBuilder#build`
3. at that point we already have access to properties, so we can simply pass a boolean flag to the `DefaultTaskExtractor` constructor?

kawamuray (changes requested):
returning once

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/DefaultTaskExtractor.java
ocadaruma: Ah, that's way better. Let me fix

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorsBuilder.java
```diff
@@ -116,15 +127,24 @@ public ProcessorsBuilder<T> thenProcess(DecatonProcessor<T> processor) {
         return thenProcess(new DecatonProcessorSupplierImpl<>(() -> processor, ProcessorScope.PROVIDED));
     }
 
-    Processors<T> build(DecatonProcessorSupplier<byte[]> retryProcessorSupplier) {
-        return new Processors<>(suppliers, retryProcessorSupplier, taskExtractor, retryTaskExtractor);
+    Processors<T> build(DecatonProcessorSupplier<byte[]> retryProcessorSupplier, ProcessorProperties properties) {
```
kawamuray, line 130: I thought that just taking an instance of `taskExtractor` as an argument of this method was ok, better to do it with "constructor" thing?

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/DefaultTaskExtractor.java
```diff
@@ -43,17 +45,33 @@ public DecatonTask<T> extract(ConsumedRecord record) {
...
-                DecatonTaskRequest taskRequest = DecatonTaskRequest.parseFrom(record.value());
-                TaskMetadata metadata = TaskMetadata.fromProto(taskRequest.getMetadata());
-                byte[] taskDataBytes = taskRequest.getSerializedTask().toByteArray();
+            // There are two cases where task metadata header is missing:
+            // 1. The task is produced by an old producer which wraps tasks in DecatonTaskRequest proto.
+            // 2. The task is produced by non-DecatonClient producer.
+            //
+            // From Decaton perspective, there is no way to distinguish between these two cases,
+            // so we need to rely on a configuration to determine how to deserialize the task.
+            if (legacyFallbackEnabledProperty.value()) {
```
kawamuray, line 54: h-m should this be dynamically reloadable? i just sense that it might be better fixed for the whole lifetime of this subscription, but idk

kawamuray (changes requested):

processor/src/main/java/com/linecorp/decaton/processor/runtime/ProcessorsBuilder.java
ocadaruma: To instantiate TaskExtractor on `build`'s caller (i.e. SubscriptionBuilder), ProcessorsBuilder needs to store `Deserializer` or `TaskExtractor` passed from user through `consuming` and SubscriptionBuilder need to instantiate appropriate resulting TaskExtractor accordingly, which is fairly complicated.

Maybe just deferring the instantiation by constructor is straightforward.

processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/DefaultTaskExtractor.java
ocadaruma: IMO we should make property reloadable unless it's difficult or there is big downside, because it can reduce the number of rolling-restart which is a certain burden and enables quick rollback.

Since upgrade-procedure might be tough already, making reloadable might be better.

The downside I can imagine is the blast radius when property change caused a problem (because property change will happen on all instances at the same time), but quick-rollback by dynamic-reloading might be more preferred than rolling restart the application I guess.

kawamuray (approved):
LGTM

PR #243, opened 2024-08-19
Add IT to check protocol migration works

ocadaruma:
- Decaton 9.0.0 contains big breaking change (#238), which needs careful upgrade procedure.
- It's better to check if the procedure works correctly with IT


kawamuray (approved):
LGTM

PR #244, opened 2024-08-19
Sonatype requires to publish using user token

ocadaruma:
- user/pass-based authentication is now deprecated on sonatype
- also included common pitfall in PGP signing


kawamuray (approved):
thanks for the update!

PR #248, opened 2025-01-27
Put back support to produce in the old DecatonTaskRequest for migration

kawamuray:
Related to: https://github.com/line/decaton/pull/238/files

In https://github.com/line/decaton/pull/238/files and https://github.com/line/decaton/pull/238/files we dropped the DecatonTaskRequest w/ online migration procedure.

However, in the apps that uses both `decaton-processor` and `decaton-client` (i.e, DecatonClient), it is impossible to migrate the app to decaton 9 or above because it inherently upgrades `decaton-client` dep by transitive dep from `decaton-processor` first.
To address this situation I put back the option to let DecatonClient keep producing tasks in older version until all processor completes upgrade. 

After this change the upgrade procedure will be:

* Step1: Upgrade decaton-processors to 9.0.0 or higher with decaton.retry.task.in.legacy.format=true, decaton.legacy.parse.fallback.enabled=true. Also set `produceInOldTaskRequest` to true in `DecatonClientBuilder` if you're using the DecatonClient in the same app.
* Step2: Upgrade decaton-clients to 9.0.0 or higher
* Step3: After all decaton-processors are upgraded, set decaton.retry.task.in.legacy.format=false, Also set `produceInOldTaskRequest` to false in `DecatonClientBuilder` if you're using the DecatonClient in the same app.
* Step4: Wait all old-format tasks and retry tasks are processed after Step3.
   * You can monitor topic's committed offset is caught-up and retry-topic's offset lag becomes 0.
* Step5: Set decaton.legacy.parse.fallback.enabled=false

### DIFF
--- a/client/src/main/java/com/linecorp/decaton/client/DecatonClientBuilder.java
+++ b/client/src/main/java/com/linecorp/decaton/client/DecatonClientBuilder.java
@@ -49,6 +49,8 @@ public class DecatonClientBuilder<T> {
     private String applicationId;
     private String instanceId;
     private KafkaProducerSupplier producerSupplier;
+    @Deprecated
+    private boolean produceInOldTaskRequest;
 
     public static class DefaultKafkaProducerSupplier implements KafkaProducerSupplier {
         @Override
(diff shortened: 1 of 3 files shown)


ocadaruma (approved):
> it inherently upgrades decaton-client dep by transitive dep from decaton-processor first

Nice catch. I overlooked this point.

LGTM

PR #251, opened 2025-05-15
Fix task parse failure during migration to v9 fixes #250

### LINKED ISSUE #250: Dropping records during decaton v9 migration due to incompatible retry task format
## Problem Overview
During migration to Decaton v9, we encountered dropping records in retry topics due to deserialization failures. The issue occurs when a processor with `decaton.retry.task.in.legacy.format=true` produces retry tasks while still having the `dt_meta` header present, causing a format mismatch when those tasks are later consumed.

Image

## Situation
* decaton-client: v9.1.0
* decaton-processor: v9.1.0 
    * `decaton.retry.task.in.legacy.format=true`
    * `decaton.legacy.parse.fallback.enabled=true`

I migrated Decaton according to Case D of [the migration guide](https://github.com/line/decaton/releases/tag/v9.0.0). After updating the clients to produce v9+ format messages, I encountered the following error logs:
```
Message:
  Dropping not-deserializable task [topic=xxx-retry partition=0 offset=1234]  

Stack trace:
  java.lang.IllegalArgumentException: com.google.protobuf.InvalidProtocolBufferException: Protocol message had invalid UTF-8.
  ...
```

## Root cause
* Decaton clients produce tasks with the `dt_meta` header if their version is v9 or higher.
* The DecatonTaskRetryQueueingProcessor produces tasks in legacy format if `decaton.retry.task.in.legacy.format=true` and retains headers, including `dt_meta`.
* The Decaton processor attempts to deserialize messages as v9+ format if the `dt_meta` header is present; otherwise, it uses the legacy format.

This leads to deserialization errors when processing tasks in the retry topic that have the `dt_meta` header and are in legacy format.

## Proposed Solution
Modify DecatonTaskRetryQueueingProcessor to remove the `dt_meta` header when `decaton.retry.task.in.legacy.format=true` to prevent the format mismatch.
The fix should be applied at: https://github.com/line/decaton/blob/ef89f5f2668cf98613070e14dcd081853e5abc90/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/DecatonTaskRetryQueueingProcessor.java#L84

ocadaruma:
## Summary
- As described in #250, current upgrade procedure for CaseD doesn't work because retry tasks might be serialized as old format when `retry.task.in.legacy.format` is true while preserving `dt_meta` if original task is produced by new decaton client
  * Existing integration test didn't catch this because the procedure written in release-note and the procedure in integration test wasn't equal... This was the mistake I overlooked.
      - Unliked CaseD, current integration test disables `retry.task.in.legacy.format` before upgrading decaton client, which works correctly
- To fix this, we should ensure `dt_meta exists iff the task is in new format` is always true, by removing `dt_meta` when producing retry task in old format

## Appendix
- I formally verified the correctness: https://github.com/line/decaton/issues/250#issuecomment-2883796173

## TODO
- Update upgrade guide in RELEASE NOTE with:
  * WARN CaseD doesn't work in Decaton <= v9.1.0 so must go with v9.1.1

### DIFF
--- a/client/src/main/java/com/linecorp/decaton/client/internal/TaskMetadataUtil.java
+++ b/client/src/main/java/com/linecorp/decaton/client/internal/TaskMetadataUtil.java
@@ -26,7 +26,7 @@ import com.google.protobuf.InvalidProtocolBufferException;
 import com.linecorp.decaton.protocol.Decaton.TaskMetadataProto;
 
 public class TaskMetadataUtil {
-    private static final String METADATA_HEADER_KEY = "dt_meta";
+    public static final String METADATA_HEADER_KEY = "dt_meta";
 
     /**
      * Write metadata to {@link Headers}
--- a/processor/src/it/java/com/linecorp/decaton/processor/ProtocolMigrationTest.java
+++ b/processor/src/it/java/com/linecorp/decaton/processor/ProtocolMigrationTest.java
@@ -158,6 +158,101 @@ public class ProtocolMigrationTest {
     }
 
     /*
+     * Check if the migration procedure doesn't cause disruption when we upgrade decaton client before disabling legacy retry.
+     * As of v9.1.0, we found that it causes message parse failure if we process upgrade like that. (https://github.com/line/decaton/issues/250)
+     */
+    @Test
+    @Timeout(30)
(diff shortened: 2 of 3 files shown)


kawamuray (approved):
ugh, thx for the fix. LGTM

PR #253, opened 2025-06-28
Fix `decaton.processor.processed.time` to measure task processing time

### LINKED ISSUE #252: Change the measurement target of `decaton.processor.processed.time`
## Question
As shown below, the metric [`decaton.processor.processed.time`](https://github.com/line/decaton/blob/v9.1.1/processor/src/main/java/com/linecorp/decaton/processor/metrics/Metrics.java#L218-L222) appears to have changed since v8. Was this intentional?

### < v8
The metric measured the total time taken by the processors to process tasks, including the time spent on scheduling.
https://github.com/line/decaton/blob/642a81d770e64549fae1dba3692d84cf42deaf25/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ProcessorUnit.java#L71-L85

### >= v8
It now appears to measure the time taken for the [`ThreadPoolExecutor#execute`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/ThreadPoolExecutor.html#execute(java.lang.Runnable)) call with the argument `() -> processTask(request)` to finish.
https://github.com/line/decaton/blob/15f4e38ca824ecef6d201c1274aa9ce6d6139ed6/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ThreadPoolSubPartitions.java#L87-L99

https://github.com/line/decaton/blob/15f4e38ca824ecef6d201c1274aa9ce6d6139ed6/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ProcessorUnit.java#L52

I believe this measures the task queuing time rather than the actual execution time of the `command`.
Ref. https://github.com/openjdk/jdk/blob/jdk-21-ga/src/java.base/share/classes/java/util/concurrent/ThreadPoolExecutor.java#L1325-L1377

## Possible Solution
If this change was unintentional, the metric should measure the execution time of `processTask(request)` directly.
Although I do not have a definitive solution, one potential approach is to modify `executor.execute(() -> processTask(request));` as follows:

```java
executor.execute(() -> {
    Timer timer = Utils.timer();
    try {
        processTask(request);
    } finally {
        metrics.processorProcessedTime.record(timer.duration());
    }
});
```

However, since this change could also affect `VirtualThreadSubPartitions`, it requires careful consideration.

emthrm:
Fix #252. I apologize if my correction is incorrect.

## Summary
- As-Is: `decaton.processor.processed.time` measured the time taken to execute [`ThreadPoolExecutor#execute`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/ThreadPoolExecutor.html#execute(java.lang.Runnable)) itself.
- To-Be: This timer measures the time taken to execute `command` (= `processTask(request)`).

### DIFF
--- a/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ThreadPoolSubPartitions.java
+++ b/processor/src/main/java/com/linecorp/decaton/processor/runtime/internal/ThreadPoolSubPartitions.java
@@ -89,12 +89,14 @@ public class ThreadPoolSubPartitions extends AbstractSubPartitions {
                                       Utils.namedThreadFactory("PartitionProcessorThread-" + scope)) {
             @Override
             public void execute(Runnable command) {
-                Timer timer = Utils.timer();
-                try {
-                    super.execute(command);
-                } finally {
-                    metrics.processorProcessedTime.record(timer.duration());
-                }
+                super.execute(() -> {
+                    Timer timer = Utils.timer();
+                    try {
+                        command.run();
+                    } finally {
+                        metrics.processorProcessedTime.record(timer.duration());
+                    }
+                });
             }
         };
     }


ocadaruma (approved):
Thank you for the quick fix! LGTM

PR #260, opened 2025-07-15
Run unintentionally ignored tests

ocadaruma:
- Found that `centraldogma`'s tests still depend on junit4, while Decaton's global test runner is already migrated to junit5 in #214 
  * Due to this, `centraldogma`'s tests were not executed at all unintentionally
- Also, `spring`'s integration tests were not executed because it isn't included in `check` task

### DIFF
--- a/centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierIntegrationTest.java
+++ b/centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierIntegrationTest.java
@@ -57,7 +57,7 @@ import com.linecorp.decaton.processor.runtime.Property;
 
 public class CentralDogmaPropertySupplierIntegrationTest {
     @RegisterExtension
-    static final CentralDogmaExtension extension = new CentralDogmaExtension() {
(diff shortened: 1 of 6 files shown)


centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierIntegrationTest.java
```diff
@@ -169,33 +177,40 @@ public void testCDRegisterSuccess() {
...
     }
 
-    @Test(timeout = 15000, expected = RuntimeException.class)
+    @Test
+    @Timeout(15)
     public void testCDRegisterTimeout() {
-        CentralDogma client = centralDogmaRule.client();
-        client.createProject(PROJECT_NAME).join();
-        CentralDogmaRepository centralDogmaRepository = spy(client.createRepository(PROJECT_NAME, REPOSITORY_NAME).join());
+        assertThrows(RuntimeException.class, () -> {
```
kawamuray, line 192: should be enclosing just the last line?

kawamuray (changes requested):

kawamuray (approved):

Issue #173, opened 2022-11-01
Provide CentralDogmaPropertySupplier that can read YAML


maediy:
The properties files on centraldogma must be written in JSON format for now.
I want to leave some comments on the properties files, but JSON doesn't support comment-out.
So, I hope that decaton-centraldogma supports YAML which supports comment-out.

ocadaruma:
related: https://github.com/line/centraldogma/issues/151

Yang-33:
I've thought about this and come up with two workarounds. I'm not sure if anyone is following the issue, but let me share them:

(1) Implement a PropertySupplier that supports YAML
decaton user can implement a YAML version of [CentralDogmaPropertySupplier](https://github.com/line/decaton/blob/021d6183d5a9b02a7fc6fbfb52cf1ceb76efc7c2/centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java#L67). By receiving the Central Dogma file as text and reading it with a YAML parser, YAML support can be achieved.
The downside is that the user must take responsibility for maintenance. Additionally, there is no guarantee that the YAML version will continue to be usable with future changes to decaton.

(2) Support JSON5 to JSON conversion in user land with CI or scripts
Central Dogma users utilize the mirroring feature with Git, especially with GitHub, which is convenient. By setting up CI or scripts to create files for decaton from JSON5, comments can be used.
For example, we can convert JSON5 files to a comment-free state with CI or scripts, place them in a different directory, and have decaton read them.
The downside is the risk (a kind of operational error?) that users might change the files intended for decaton after conversion.

I hope this is helpful.

It seems the use of Central Dogma with decaton is very simple. Even without using Central Dogma's JSON support, we  can achieve this issue by receiving the file as a text file and using an object mapper to interpret it as JSON or YAML (probably). I've known and used decaton over three years(i love decaton), and I've often wanted to add notes or comments about ignored keys to the decaton files in Central Dogma to explain history or caution.

PR #262, opened 2025-07-23
Support yaml partially

Yang-33:
As far as I know and one of decaton users, inline comments are often necessary for `CONFIG_IGNORE_KEYS`, `CONFIG_PARTITION_CONCURRENCY`, and `CONFIG_MAX_PENDING_RECORDS`.

Using YAML allows us to add inline comment in central dogma decaton property file.
Decaton's internal parser is Jackson, which seems to partially support YAML. By leveraging this, YAML can be supported with relatively few changes. 

Of course, it’s not complete. While YAML tags, anchors, and aliases cannot be used, we can include comments in the configuration file. Practically speaking, this seems to be sufficient...

If maintainers don't want to use YAML with parital support, there are some Jackson options that support JSON5 to support comment in json file.
- https://github.com/FasterXML/jackson-core/wiki/JsonReadFeatures#support-for-non-standard-json-content-comments

Resolve https://github.com/line/decaton/issues/173

### DIFF
--- a/docs/dynamic-property-configuration.adoc
+++ b/docs/dynamic-property-configuration.adoc
@@ -82,7 +82,7 @@ public class CentralDogmaSupplierMain {
 
 == Use YAML instead of JSON
 
-From decaton v10, you can store the property file in YAML as well as JSON.
+From decaton v9.4.0, you can store the property file in YAML as well as JSON.
 Nothing changes in your code except the file‐name extension.
 
 Note you cannot use YAML's tag, anchor, or alias features. Just you can add comment in the file.


ocadaruma:
Hi, thank you for the patch and sorry for the late review.

Could you resolve the conflict, for the ease of review?

Yang-33:
rebased!

centraldogma/src/test/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplierIntegrationTest.java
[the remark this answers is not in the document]
Yang-33: and I know CI( CI /test (java8 ) failed. Let me fix later if this is acceptable

centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
```diff
@@ -103,6 +117,21 @@ public CentralDogmaPropertySupplier(CentralDogmaRepository centralDogmaRepositor
         } catch (TimeoutException e) {
             throw new RuntimeException(e);
         }
+
+        rootWatcher.watch(node -> {
+            node.fields().forEachRemaining(entry -> {
```
ocadaruma, line 122: In contrast to current approach, new code sets the watch against entire file and iterates through all properties in the file.
Do you have any thoughts if it will cause some downside comparing to current approach?

ocadaruma (commented):
Left one early question, but overall strategy looks good.

Please go ahead and complete the implementation. Thanks for the patch!

centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
Yang-33: oops, we might want to avoid using `node.fields()` in this change, and instead explicitly check properties based on `ProcessorProperties#PROPERTY_DEFINITIONS`.
There’s no benefit in processing keys that Decaton doesn’t care about. Let me modify this point, though it's not main part for your question...

ocadaruma: Thank you for the detailed explanation including the implementation of CD!

I see, then the strategy looks good.

docs/dynamic-property-configuration.adoc
```diff
@@ -80,6 +80,53 @@ public class CentralDogmaSupplierMain {
 }
 ----
 
+== YAML Support
```
Yang-33, line 83: Let me update document after https://github.com/line/decaton/pull/263 is merged into master. We can use jsonschema for YAML too.

centraldogma/src/main/java/com/linecorp/decaton/centraldogma/CentralDogmaPropertySupplier.java
```diff
@@ -103,6 +119,21 @@ public CentralDogmaPropertySupplier(CentralDogmaRepository centralDogmaRepositor
         } catch (TimeoutException e) {
             throw new RuntimeException(e);
         }
+
+        rootWatcher.watch(node -> {
+            for (PropertyDefinition<?> definition : PROPERTY_DEFINITIONS) {
+                DynamicProperty<?> p = cachedProperties.get(definition.name());
+                if (p != null && node.has(definition.name())) {
```
Yang-33, line 126: 🙅This is wrong. 
Actually, the PKQ-related settings should also have been updated, but this didn’t work... sorry about that.
We need to retrieve the key–value from the JSON node based on the cached property names. I’ll fix it.

Yang-33: resolved

docs/dynamic-property-configuration.adoc
Yang-33: resolved

centraldogma/src/main/java/com/linecorp/decaton/centraldogma/DecatonPropertyFileFormat.java
ocadaruma, line 1: This class doesn't need to be exposed to users, so let's move to `internal` package (refs: https://github.com/line/decaton/blob/master/VERSIONING.md#public-apis)

JsonFormat, YamlFormat as well

ocadaruma (commented):
Sorry for the late review.
Left only one minor comment. overall LGTM!

ocadaruma:
Oh is this breaking change? I thought no

PR #271, opened 2026-03-12
Call processingReturn and register processingCompletion even if task fails to process exceptionally 

zzhen0v0:
# Why
The tracing handle does not seem to handle the exceptional path if an exception is thrown from processing, which contradicts the handler's promises. It might result in trace leaks.
1. [processingCompletion](https://github.com/line/decaton/blob/master/processor/src/main/java/com/linecorp/decaton/processor/tracing/TracingProvider.java#L38-L43)
2. [processingReturn](https://github.com/line/decaton/blob/master/processor/src/main/java/com/linecorp/decaton/processor/tracing/TracingProvider.java#L57-L65)

# What
This change is to move processingReturn to be called, and processingCompletion to be registered around the finally block, to guarantee no trace leak even in the exceptional path.


m50d (approved):

mauhiz (approved):

ocadaruma (approved):
Sorry for the delay, LGTM.
Thank you!!

PR #272, opened 2026-03-27
Fix shadow build failure

ocadaruma:
Currently, `./gradlew :benchmark:shadowJar` fails with below error:
```
* What went wrong:
Execution failed for task ':benchmark:shadowJar'.
> Unsupported class file major version 65
```

Suspecting this https://github.com/GradleUp/shadow/issues/894.
Not 100% sure why and when this error started happen though, let's just upgrade.

Confirmed locally the issue is fixed.

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -76,7 +76,7 @@ public class CustomRunner implements Runner {
 
 #### Build a shadowJar of the custom runner:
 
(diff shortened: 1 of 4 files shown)


Yang-33 (approved):
thank you so much!

(record) https://github.com/johnrengelman/shadow  is moved to https://github.com/GradleUp/shadow so it's just upgrading version

PR #277, opened 2026-05-09
Use version catelogs in Gradle

Yang-33:
Version catalogs is a modern way to declare versions. https://docs.gradle.org/current/userguide/version_catalogs.html

This change migrates the current version declarations to version catalogs. The scope of this change is only aliases, so it does not affect publishing or dependency resolution.

Version catalog can also manage plugin versions, but I did not introduce that yet because the current `build.gradle` style cannot use it. After we finish cleaning up the Gradle setup, we may be able to introduce it.

I confirmed that this does not change version resolution or the published JARs as follows:

1. Locally publish from the master branch and get the POM for the corresponding JAR.
2. Do the same from this PR.
3. Compare them and confirm they are identical.

### DIFF
--- a/benchmark/build.gradle
+++ b/benchmark/build.gradle
@@ -1,25 +1,22 @@
 ext {
     publishOnlyLocal = true
     shadeAllDependencies = true
-
-    picoVersion = "4.2.0"
-    jacksonVersion = "2.13.4"
 }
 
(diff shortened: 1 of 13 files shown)


ocadaruma (approved):
Sorry for the delay, LGTM thanks!

PR #278, opened 2026-05-09
Upgrade Gradle to v9.5.1 from v8.5

Yang-33:
The current Gradle version is from two years ago. Let’s upgrade it to v9. The guide is here:
[https://docs.gradle.org/current/userguide/upgrading_major_version_9.html](https://docs.gradle.org/current/userguide/upgrading_major_version_9.html)

There were almost no difficult parts.

The main points are:

1. Some old plugin versions caused build failure because they do not support Gradle v9, so this change upgrades the plugin versions.
2. Test tasks now require stricter classpath and directory configuration. This affected the integration tests. To be safe,  this change confirms that the number of executed tests after this change matches the current master branch, 215 unit tests + 41 itests.
    - https://docs.gradle.org/current/userguide/upgrading_major_version_9.html#test_tasks_may_no_longer_execute_expected_tests
3. `junit-platform-launcher`, which was previously used implicitly, now needs to be declared explicitly, so this change adds it.
    - https://docs.gradle.org/current/userguide/upgrading_version_8.html#test_framework_implementation_dependencies

A local publish comparison was also performed, confirming no major impact on the JARs.


build.gradle
```diff
@@ -83,10 +85,19 @@ subprojects {
         }
     }
 
+    sourceSets.create('it') {
+        compileClasspath += sourceSets.main.output
+        runtimeClasspath += sourceSets.main.output
+        compileClasspath += sourceSets.testFixtures.output
+        runtimeClasspath += sourceSets.testFixtures.output
+    }
```
Yang-33, line 93: The order here is important.

Since Gradle 8.14.4, accessing reserved configuration names derived from `sourceSet` before creating `sourceSets` has been deprecated. In Gradle 9, it becomes an error.

Without this change, the following error occurs:

```text
Configuration with name 'itImplementation' not found.
```

.github/workflows/ci.yml
```diff
@@ -16,34 +14,37 @@ jobs:
...
     - name: Setup java
-      uses: actions/setup-java@v4
+      uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5.2.0
       with:
         distribution: temurin
         java-version: |
           ${{ matrix.java }}
           21
+    - name: Setup Gradle
+      uses: gradle/actions/setup-gradle@0723195856401067f7a2779048b490ace7a47d7c # v5.0.2
```
Yang-33, line 26: `eskatos/gradle-command-action` has been moved to `gradle/actions/setup-gradle` finally.
- https://github.com/eskatos/gradle-command-action -> https://github.com/gradle/gradle-build-action -> https://github.com/gradle/actions/tree/main/setup-gradle

Yang-33:
it's ready to be reviewed after https://github.com/line/decaton/pull/277 is merged into master branch~

Yang-33:
(rebased but no diff: https://github.com/line/decaton/compare/7260792415201f34bf506e23be0b2fd013d5bc88..16b97ffd679a1dcbcd32bb1fd790ecedcfa99ef0)

Yang-33:
It's ready to be reviewed!

build.gradle
```diff
@@ -226,7 +232,7 @@ subprojects {
                 }
             }
             signing {
-                required { isReleaseVersion && gradle.taskGraph.hasTask("publish") }
+                required = isReleaseVersion && gradle.taskGraph.hasTask("publish")
```
ocadaruma: Is removing `{}` safe here?
In my understanding, this makes the evaluation to be eager, then evaluated at configuration timing before graph is finalized, so it may bypass the check and may be evaluated as `false` unintentionally

ocadaruma (commented):
Thanks!
Left one comment

build.gradle
Yang-33: I didn't know that, thank you for pointing out this 🙏 

https://docs.gradle.org/9.5.0/javadoc/org/gradle/plugins/signing/SigningExtension.html#setRequired(java.lang.Object)

Image

ocadaruma (approved):

